PR #2, opened 2026-02-21
Active learning loop

pedromsferraz-mila:
This PR introduces the core active-learning architecture (loop + abstract interfaces + dummy implementations) across acquisition, sampler, selector, surrogate, dataset, oracle, budget, and shared types.

It adds multi-fidelity capabilities (including composite oracle routing, fidelity confidences, and budget/cost-aware behavior), plus related refactors for flexibility (iterable datasets, optional observations in acquisition updates, and improved sampler/selector behavior).

It also standardizes documentation by converting docstrings to typed NumPy format and adds comprehensive tests.


src/activelearning/active_learning.py
```diff
@@ -0,0 +1,103 @@
...
+    """Execute the active learning loop with budget constraints.
+
+    Iteratively: (1) fits surrogate on current data, (2) samples candidates,
+    (3) selects candidates to label, (4) queries oracle and adds observations.
+    Stops when budget is exhausted or no affordable candidates remain.
+
+    Parameters
+    ----------
+        dataset : Dataset
+            Dataset for storing and retrieving observations.
```
alexhernandezgarcia: This comment applies more generally than this line but I make it here as an example.

I believe that the indentation of the list of parameters (and returned variables, etc.) should be reduced, as per the standard in the Numpy style Python docstrings. The indentation of the variable name should be the same of `Parameters` and `----------`. The indentation of the description of the variable is one level higher.

See examples [here](https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html).

src/activelearning/oracle/multi_fidelity_oracle.py
```diff
@@ -0,0 +1,132 @@
...
+        fidelity_configs: dict[int, dict[str, Any]],
+    ) -> None:
+        """Initialize MultiFidelityOracle with fidelity configurations.
+
+        Raises
+        ------
+            ValueError
+                If fidelity is not an integer or required keys are missing.
+        """
+        # Validate fidelity configs
```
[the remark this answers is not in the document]
pedromsferraz-mila: Input validation will be done by the config parser, when we implement it.

tests/selector/test_cost_aware_selector.py
```diff
@@ -0,0 +1,253 @@
...
+    # Classic knapsack counter-example
+    candidates = [Candidate(x=0), Candidate(x=1)]
+    acquisition = Mock()
+    acquisition.return_value = [10.0, 9.0]  # Candidate 0 slightly better utility
+
+    def cost_fn(c):
+        return [6.0, 5.0]  # Candidate 0 costs more
+
+    # Budget 10.0 - greedy picks candidate 0 (ratio 10/6=1.67)
+    # But candidate 1 (ratio 9/5=1.8) is actually better and would fit twice
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [47a75f7](https://github.com/milaforscience/activelearning/pull/2/commits/47a75f7b1175bf98dfa05078b1b90090f3a1bfb4)

src/activelearning/selector/cost_aware_selector.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Greedily select candidates until budget exhausted
+        selected = []
+        budget_used = 0.0
+
+        for _, idx in ratios:
+            candidate_cost = costs[idx]
+            if budget_used + candidate_cost <= round_budget:
+                selected.append(candidates[idx])
+                budget_used += candidate_cost
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [5aad76a](https://github.com/milaforscience/activelearning/pull/2/commits/5aad76a6e4c9203363d08235182c7c18dde631de)

src/activelearning/oracle/multi_fidelity_oracle.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        observations = []
+        for candidate in candidates:
+            score_fn = self.fidelity_configs[candidate.fidelity]["score_fn"]
+            observation = Observation(
+                x=candidate.x,
+                y=score_fn(candidate.x),
+                fidelity=candidate.fidelity,
+            )
+            observations.append(observation)
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [ab6a3a0](https://github.com/milaforscience/activelearning/pull/2/commits/ab6a3a00156190a4f72bd9bd96113779bb432ebc)

tests/oracle/test_multi_fidelity_oracle.py
```diff
@@ -0,0 +1,364 @@
...
+
+    assert observations[1].x == 20
+    assert observations[1].y == 20.5
+    assert observations[1].fidelity == 1
+
+    assert observations[2].x == 30
+    assert observations[2].y == 30.0
+    assert observations[2].fidelity == 0
+
+
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [ab6a3a0](https://github.com/milaforscience/activelearning/pull/2/commits/ab6a3a00156190a4f72bd9bd96113779bb432ebc)

src/activelearning/sampler/pool_score_sampler.py
```diff
@@ -0,0 +1,63 @@
...
+            raise ValueError("Acquisition function is required for PoolScoreSampler.")
+
+        scores = acquisition(self.candidate_pool)
+
+        # Apply softmax to convert scores to valid probabilities
+        weights = torch.softmax(torch.tensor(scores), dim=0).tolist()
+
+        return random.choices(
+            population=list(self.candidate_pool), weights=weights, k=self.num_samples
+        )
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [c02e502](https://github.com/milaforscience/activelearning/pull/2/commits/c02e50252ca3044894b1e97aafafeac3f0c0773a)

src/activelearning/oracle/multi_fidelity_oracle.py
```diff
@@ -0,0 +1,132 @@
...
+                If fidelity is not an integer or required keys are missing.
+        """
+        # Validate fidelity configs
+        for fidelity, config in fidelity_configs.items():
+            if not isinstance(fidelity, int):
+                raise ValueError(f"Fidelity must be int, got {type(fidelity)}")
+            if "cost_per_sample" not in config:
+                raise ValueError(
+                    f"Missing 'cost_per_sample' in config for fidelity {fidelity}"
+                )
```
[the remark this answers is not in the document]
pedromsferraz-mila: Input validation will be done by the config parser, when we implement it.

tests/test_active_learning.py
```diff
@@ -0,0 +1,129 @@
...
+    active_learning(
+        dataset=dataset,
+        surrogate=surrogate,
+        acquisition=acquisition,
+        sampler=sampler,
+        selector=selector,
+        oracle=oracle,
+        budget=budget,
+    )
+    assert surrogate.fidelity_confidences == oracle.get_fidelity_confidences()
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [60887b1](https://github.com/milaforscience/activelearning/pull/2/commits/60887b19c03615c5075e59d6b5ebc7cb729eef41)

src/activelearning/active_learning.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+        # Query oracle to obtain total cost for the samples
+        costs = oracle.get_costs(selected_samples)
+        total_cost = sum(costs)
+
+        # Check if we can afford this query before consuming budget
+        if not budget.can_afford(total_cost):
+            # Budget exhausted - stop iteration
+            break
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [60887b1](https://github.com/milaforscience/activelearning/pull/2/commits/60887b19c03615c5075e59d6b5ebc7cb729eef41)

src/activelearning/oracle/multi_fidelity_oracle.py
```diff
@@ -0,0 +1,132 @@
...
+            if "score_fn" not in config:
+                raise ValueError(
+                    f"Missing 'score_fn' in config for fidelity {fidelity}"
+                )
```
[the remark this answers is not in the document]
pedromsferraz-mila: Input validation will be done by the config parser, when we implement it.

src/activelearning/active_learning.py
pedromsferraz-mila: Good point, fixed! [9728600](https://github.com/milaforscience/activelearning/pull/2/commits/9728600902d2621a1b992995f4eba37afdcfca9b)

src/activelearning/dataset/dataset.py
```diff
@@ -0,0 +1,90 @@
+from abc import ABC, abstractmethod
+from typing import Sequence, Iterable
+
+from activelearning.utils.types import Observation
+
+
+class Dataset(ABC):
```
alexhernandezgarcia, line 7: No need for `__init__()`?

src/activelearning/dataset/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+            - A DataLoader or generator for large datasets (creates fresh iterator each call)
+            - Any iterable that can be traversed to access observations
+
+        Notes
+        -----
+            Concrete implementations should ensure this method is efficient to call
+            multiple times per active learning iteration. For example:
+            - Returning a list reference is O(1)
+            - Creating a new DataLoader is typically fast (just wraps existing data)
+            - Avoid expensive recomputation on each call
```
alexhernandezgarcia: Reduce indentation one level.
```suggestion
        Concrete implementations should ensure this method is efficient to call
        multiple times per active learning iteration. For example:
        - Returning a list reference is O(1)
        - Creating a new DataLoader is typically fast (just wraps existing data)
        - Avoid expensive recomputation on each call
```

src/activelearning/dataset/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+            ```python
+            def get_observations_iterable(self):
+                return self._records  # Returns same list each time
+            ```
+
+            DataLoader implementation (fresh iterator each call):
+            ```python
+            def get_observations_iterable(self):
+                return DataLoader(self._dataset)  # New DataLoader each time
+            ```
```
alexhernandezgarcia: Reduce indentation one level.
```suggestion
        Simple list implementation (cheap multiple calls):
        ```python
        def get_observations_iterable(self):
            return self._records  # Returns same list each time
        ```

        DataLoader implementation (fresh iterator each call):
        ```python
        def get_observations_iterable(self):
            return DataLoader(self._dataset)  # New DataLoader each time
        ```
```

src/activelearning/dataset/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+        Returns
+        -------
+        best_candidates : list[Observation]
+            List of top-k observations sorted by y value (descending).
+            Returns empty list if no observations exist.
+
+        Notes
+        -----
+            Filters out observations with None y values.
+            Assumes y values support comparison operations.
```
alexhernandezgarcia: Reduce indentation one level.
```suggestion
        Filters out observations with None y values.
        Assumes y values support comparison operations.
```

src/activelearning/dataset/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+
+            DataLoader implementation (fresh iterator each call):
+            ```python
+            def get_observations_iterable(self):
+                return DataLoader(self._dataset)  # New DataLoader each time
+            ```
+        """
+        pass
+
+    def get_best_candidates(self, k: int = 1) -> list[Observation]:
```
alexhernandezgarcia, line 59: I would not change anything now, but I wanted to note that the notions of "best" and "better" may change from problem to problem. For example, higher may not always be better. Also, the `Observation` types supports categorical labels too.

I guess one way to deal with this would be to leave it up to sub-classes, but in that case it would be good to add a note here.

src/activelearning/dataset/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+        Notes
+        -----
+            Filters out observations with None y values.
+            Assumes y values support comparison operations.
+        """
+        observations = list(self.get_observations_iterable())
+        if not observations:
+            return []
+        valid_obs = [o for o in observations if o.y is not None]
+        sorted_records = sorted(valid_obs, key=lambda r: r.y, reverse=True)
```
alexhernandezgarcia: Will this not create multiple (unnecessary) copies of the data, which might take time and space?

src/activelearning/active_learning.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    initial_budget = budget.available_budget
+    num_rounds = 0
+
+    # Set fidelity confidences in surrogate before starting the loop
+    surrogate.set_fidelity_confidences(oracle.get_fidelity_confidences())
+
+    while budget.available_budget > 0:
+        # Note: get_observations_iterable() is called multiple times to avoid
+        # consuming the same iterable across multiple consumers
```
alexhernandezgarcia: I am wondering whether this could introduce a risk of constructing different data sets for each component. For example, if the iterable is a DataLoader with some random operations.

src/activelearning/surrogate/surrogate.py
```diff
@@ -0,0 +1,97 @@
+from abc import ABC, abstractmethod
+from typing import Iterable, Mapping, Sequence, Any
+
+from activelearning.utils.types import Candidate, Observation
+
+
+class Surrogate(ABC):
+    """Abstract surrogate interface used by acquisitions and the active learning loop.
+
+    Surrogate models approximate the true evaluation function based on
```
alexhernandezgarcia:
```suggestion
    Surrogate models approximate the oracles based on
```

src/activelearning/surrogate/surrogate.py
```diff
@@ -0,0 +1,97 @@
...
+    observed data, enabling efficient candidate evaluation.
+
+    Notes
+    -----
+        The predict() method is optional. Surrogates that only work with
+        specific acquisition functions (e.g., those using internal posterior
+        representations) may not implement general prediction.
```
alexhernandezgarcia: Reduce indentation one level.
```suggestion
    The predict() method is optional. Surrogates that only work with
    specific acquisition functions (e.g., those using internal posterior
    representations) may not implement general prediction.
```

src/activelearning/surrogate/surrogate.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+
+    @abstractmethod
+    def fit(self, observations: Iterable[Observation]) -> None:
+        """Fit the surrogate model to observations.
+
+        This method may be called multiple times during the active learning loop
+        as new observations are collected. Implementations are not required to
+        support incremental learning (full retraining is acceptable).
```
alexhernandezgarcia:
```suggestion
        This method is called at each iteration of the active learning loop,
        as new observations are collected. Implementations are not required to
        support incremental learning (full retraining is acceptable).
```

src/activelearning/surrogate/surrogate.py
```diff
@@ -0,0 +1,97 @@
...
+        observations : Iterable[Observation]
+            Iterable of observations to train on. May be a Sequence
+            (list, tuple) for small datasets or a one-pass iterable (DataLoader)
+            for large datasets.
+
+        Notes
+        -----
+            Implementations should validate their input requirements:
+            - If you need len() or multiple passes, materialize to list or assert Sequence
+            - If you can handle streaming data, consume the iterable directly
```
alexhernandezgarcia: Reduce indentation one level.
```suggestion
        Implementations should validate their input requirements:
        - If you need len() or multiple passes, materialize to list or assert Sequence
        - If you can handle streaming data, consume the iterable directly
```

src/activelearning/surrogate/surrogate.py
```diff
@@ -0,0 +1,97 @@
+from abc import ABC, abstractmethod
+from typing import Iterable, Mapping, Sequence, Any
+
+from activelearning.utils.types import Candidate, Observation
+
+
+class Surrogate(ABC):
```
alexhernandezgarcia, line 7: No need for `__init__()`?

src/activelearning/surrogate/dummy_surrogate.py
```diff
@@ -0,0 +1,69 @@
...
+    """Minimal in-memory surrogate using cached observations.
+
+    Stores observations keyed by (x, fidelity). When predicting, returns the
+    observed value for known candidates and the global mean for unseen ones.
+    Useful for testing and baseline comparisons.
+
+    Notes
+    -----
+        This surrogate requires a Sequence (supports len() and multiple iterations).
+        Automatically materializes iterables to lists as needed.
```
alexhernandezgarcia: Reduce indentation one level.
```suggestion
    This surrogate requires a Sequence (supports len() and multiple iterations).
    Automatically materializes iterables to lists as needed.
```

src/activelearning/surrogate/dummy_surrogate.py
```diff
@@ -0,0 +1,69 @@
+from typing import Any, Iterable, Mapping, Sequence
+
+from activelearning.surrogate.surrogate import Surrogate
+from activelearning.utils.types import Candidate, Observation
+
+
+class DummySurrogate(Surrogate):
```
alexhernandezgarcia: How about using a name that explicitly indicates its behaviour, for example `DummySurrogateMean` or `DummyMeanSurrogate`?

src/activelearning/surrogate/dummy_mean_surrogate.py
```diff
@@ -0,0 +1,69 @@
...
+        ----------
+        observations : Iterable[Observation]
+            Iterable of observations to cache. Will be materialized
+            to a Sequence if needed for len() and multiple iteration.
+        """
+        # Materialize to list if not already a Sequence
+        if not isinstance(observations, Sequence):
+            observations = list(observations)
+
+        if not observations:
```
alexhernandezgarcia: This is meant to catch an empty list? If so, how about being more explicit: `if len(observations) == 0`?

Why or when would that happen? If so, is it a good idea to return a mean of 0.0? That may lead to silent issues.

src/activelearning/surrogate/dummy_mean_surrogate.py
```diff
@@ -0,0 +1,69 @@
...
+            Known candidates return cached values with low std (0.1),
+            unknown candidates return global mean with high std (1.0).
+        """
+        means = []
+        stds = []
+        for candidate in candidates:
+            key = (candidate.x, candidate.fidelity)
+            if key in self._model:
+                means.append(self._model[key])
+                stds.append(0.1)
```
alexhernandezgarcia: Nit-picky comment, but these hard-coded values could be attributes of the class :P

src/activelearning/selector/selector.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+
+    @abstractmethod
+    def __call__(
+        self,
+        candidates: Sequence[Candidate],
+        acquisition: Optional[Any] = None,
+        cost_fn: Optional[Callable[[Sequence[Candidate]], list[float]]] = None,
+        round_budget: Optional[float] = None,
+        **kwargs: Any,
```
alexhernandezgarcia: Is `Any` the right type for `kwargs`? Does it not need to be a dict?

src/activelearning/selector/score_selector.py
```diff
@@ -0,0 +1,53 @@
+from typing import Any, Callable, Optional, Sequence
+
+from activelearning.acquisition.acquisition import Acquisition
+from activelearning.selector.selector import Selector
+from activelearning.utils.types import Candidate
+
+
+class ScoreSelector(Selector):
```
alexhernandezgarcia: How about calling this something like `TopKAcquisitionSelector`?

src/activelearning/selector/cost_aware_selector.py
```diff
@@ -0,0 +1,93 @@
...
+        if not candidates:
+            return []
+
+        # Get utilities and costs for all candidates
+        utilities = acquisition(candidates)
+        costs = cost_fn(candidates)
+
+        # Reject negative costs and calculate bang-for-buck ratios
+        ratios = []
+        for i, (utility, cost) in enumerate(zip(utilities, costs)):
```
alexhernandezgarcia: This is a fixation of mine but I try to avoid using single-character variable names, which sometimes become hard to interpret. Not the case here, but just for house keeping :) Suggesting the same change where it is used.
```suggestion
        for idx, (utility, cost) in enumerate(zip(utilities, costs)):
```

src/activelearning/selector/cost_aware_selector.py
```diff
@@ -0,0 +1,93 @@
...
+            if cost < 0:
+                raise ValueError("Cost function returned a negative cost.")
+            if cost == 0:
+                # Infinite value for zero cost - select first
+                ratio = float("inf")
+            else:
+                ratio = utility / cost
+            ratios.append((ratio, i))
```
alexhernandezgarcia:
```suggestion
            ratios.append((ratio, idx))
```

src/activelearning/selector/cost_aware_selector.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # Greedily select candidates until budget exhausted
+        selected = []
+        budget_used = 0.0
+
+        for _, idx in ratios:
+            candidate_cost = costs[idx]
+            if budget_used + candidate_cost <= round_budget:
+                selected.append(candidates[idx])
+                budget_used += candidate_cost
```
alexhernandezgarcia, line 87: Not a strong suggestion, but I guess this would not harm and could save time:
```suggestion
                budget_used += candidate_cost
                if budget_used > round_budget:
                    break
```

src/activelearning/sampler/pool_score_sampler.py
```diff
@@ -0,0 +1,63 @@
...
+            List of sampled candidates weighted by softmax probabilities,
+            without replacement.
+        """
+        if self.num_samples >= len(self.candidate_pool):
+            return list(self.candidate_pool)
+
+        if acquisition is None:
+            raise ValueError("Acquisition function is required for PoolScoreSampler.")
+
+        scores = acquisition(self.candidate_pool)
```
alexhernandezgarcia: Just noticing that the outputs of the `acquisition()` are given different names in different parts of the code (`scores`, `values`, `utilities`...). I would try to harmonise this. I don't have a strong preference, but I would say that we already use "scores" in the GFlowNet repo for the proxy values; and it's quite generic. "Values" is quite generic too. "Utilities" may be better. And I would consider the more explicit `acqvalues` or `acq_values` or `values_acq`.

src/activelearning/sampler/pool_score_sampler.py
```diff
@@ -0,0 +1,63 @@
+import torch
+
+from typing import Any, Iterable, Optional, Sequence
+from activelearning.acquisition.acquisition import Acquisition
+from activelearning.sampler.sampler import Sampler
+from activelearning.utils.types import Candidate, Observation
+
+
+class PoolScoreSampler(Sampler):
+    """Samples from a candidate pool weighted by acquisition scores using softmax.
```
alexhernandezgarcia: Say explicitly "without replacement"?

src/activelearning/active_learning.py
```diff
@@ -0,0 +1,107 @@
...
+        total_cost = sum(costs)
+
+        # Check if we can afford this query before consuming budget
+        if not budget.can_afford(total_cost):
+            # Budget exhausted - stop iteration
+            break
+
+        # Consume budget and query oracle for new observations, then add to dataset
+        budget.consume(total_cost)
+        new_observations = oracle.query(selected_samples)
```
alexhernandezgarcia, line 101: I suppose you have already thought of this, but just in case: I am wondering whether it could make sense to pass the budget to the oracle, so as to minimise the possibility of accidentally calling expensive oracles in random parts of the code.

src/activelearning/oracle/composite_oracle.py
```diff
@@ -0,0 +1,199 @@
...
+    sub_oracles : list[Oracle]
+        List of oracle instances to delegate to.
+    """
+
+    def __init__(self, sub_oracles: list[Oracle]) -> None:
+        if not sub_oracles:
+            raise ValueError("CompositeOracle requires at least one sub-oracle")
+        self._sub_oracles = sub_oracles
+
+    def get_fidelity_confidences(self) -> dict[int, float]:
```
alexhernandezgarcia, line 27: Would it be possible to add a few words for the Returns part of docstring?

alexhernandezgarcia (commented):
Nice work! This is a super solid base for the library! I have dropped quite a bunch of comments and minor suggestions but I don't recall anything major.

src/activelearning/dataset/dataset.py
pedromsferraz-mila: Not necessary! This class serves as an abstract base class (ABC) that defines the interface for all dataset implementations. Its purpose is to specify which methods concrete subclasses must implement (i.e., `add_observations` and `get_observations_iterable`).

While `get_best_candidates` provides a default implementation for convenience, there's no state to initialize at this abstract level. Concrete subclasses will define their own storage mechanisms (e.g., lists, dataloaders) and initialize them in their own `__init__` methods as needed.

src/activelearning/dataset/dataset.py
pedromsferraz-mila: Good point! I've moved the implementation to `DummyDataset` and updated the docstrings to clarify:

1. In the abstract `Dataset` class: Added a note that "best" is problem-specific (maximization vs minimization vs categorical) and subclasses must define their own ranking logic
2. In `DummyDataset`: Documented that this implementation assumes a maximization problem and suggests overriding for other problem types

This makes it clearer that the notion of "best" varies by problem, and concrete implementations should provide appropriate logic for their specific use case. 8fce14d3adffeadb7fb6dcfe098dd1ff0861cc07

src/activelearning/dataset/dataset.py
pedromsferraz-mila: Good point! Initially, I thought of adding this as just a dummy implementation, so I wanted to make sure the code is easy to read and didn't mind the overhead. However, I believe it could be useful to have a list dataset implemented, so I've changed its name from `DummyDataset` to `ListDataset` and optimized the implementation to avoid unnecessary copies:
- Changed from list comprehension to generator expression
- Replaced full sort with `heapq.nlargest()` to use a heap instead, which has complexity O(n log k)

8fce14d3adffeadb7fb6dcfe098dd1ff0861cc07

src/activelearning/active_learning.py
pedromsferraz-mila: Good point! I've made some changes to address this:

   1. `get_observations_iterable()` is now called once per round in `active_learning.py`, and the same object is passed to `surrogate.fit`, `acquisition.update`, and `sampler.sample`, so there's no opportunity for divergence at the call site.
   2. The Dataset contract now explicitly requires that the returned `Iterable` supports consistent multiple iterations: calling `iter()` on it multiple times must yield the same sequence. This means a `DataLoader`-backed implementation must pin any randomness (e.g. pre-generate the shuffled index 
  permutation) when `get_observations_iterable()` is called, rather than re-sampling on each pass. The docstring documents this requirement and shows an example.

This keeps the fix minimal (no new classes or memory overhead) while making the consistency contract explicit. [30640cf](https://github.com/milaforscience/activelearning/pull/2/commits/30640cfd2750c0f089c418f015d5a682aa7616de)

src/activelearning/surrogate/surrogate.py
pedromsferraz-mila: Same as for the `Dataset` class, this class serves as an abstract base class (ABC) that defines the interface for all surrogate implementations. Its purpose is to specify which methods concrete subclasses must implement, so no need to provide an init method here.

src/activelearning/dataset/dataset.py
alexhernandezgarcia: Perfect, I agree with the decisions and changes made in https://github.com/milaforscience/activelearning/commit/8fce14d3adffeadb7fb6dcfe098dd1ff0861cc07, thanks!

src/activelearning/selector/selector.py
pedromsferraz-mila: Good catch! I had originally included `**kwargs` as a catch-all for future needs, but it's unnecessary as `__call__` already has all the parameters it needs. Any selector-specific configuration should be passed through the constructor of the concrete subclass instead. db9467f6180b91d602cbdaa7c7c21d90ad16de6c

src/activelearning/active_learning.py
alexhernandezgarcia: Great, thanks! I agree with the decisions and changes in https://github.com/milaforscience/activelearning/pull/2/commits/30640cfd2750c0f089c418f015d5a682aa7616de. Also the code in the active learning loop becomes a bit less verbose for the same price! :)

src/activelearning/dataset/dataset.py
alexhernandezgarcia: Perfect, thanks!

src/activelearning/selector/cost_aware_selector.py
pedromsferraz-mila: Good point, fixed! 54d56402029fda8420d0aa3bc914c166f2ce146d

src/activelearning/surrogate/dummy_surrogate.py
pedromsferraz-mila: Yes, good point! Fixed: 4551632709710fb75b81cd230cee2594ab3457d0

However, I didn't intend to keep the dummy surrogate in the codebase long-term. It's just a temporary placeholder so we can run end-to-end tests on the active learning loop until the actual surrogates are ready. We can remove it once we have real models in place.

src/activelearning/surrogate/dummy_mean_surrogate.py
pedromsferraz-mila: Good point, as `observations` is already a list in this case, these are equivalent. Updated: abfcfea8a3578faf6185193f94895be310365df8

As for when this happens: at the start of the active learning loop, the dataset may be empty (no initial observations), so fit gets called with an empty list before any samples have been collected.

The 0.0 mean is intentional for this dummy implementation, since all unseen candidates also fall back to `_mean_score`, returning 0.0 means every candidate gets the same score, which will produce uniform (uninformed) sampling. The goal was to signal "no information yet" rather than introducing a bias. A more realistic surrogate would handle this differently (e.g., raise an error or use a prior), but for testing purposes this is a safe, predictable default.

src/activelearning/surrogate/dummy_mean_surrogate.py
pedromsferraz-mila: Good point, fixed! f2764b5388ef057f8e25cd41ff80352c77483024

src/activelearning/selector/score_selector.py
pedromsferraz-mila: Good idea, that's much more clear! Fixed: 5035ac5a8741c5e662843d5a6582a3b0930d05a2

src/activelearning/sampler/pool_score_sampler.py
pedromsferraz-mila: Good point, fixed: 5467d68642d784e626d372519a8df3093b8905c6

src/activelearning/sampler/pool_score_sampler.py
pedromsferraz-mila: Good point! I think "acquisition values" is the most explicit, updated docstrings and variable names accordingly: 6e22d732a78a7508f48062508eb66b6e553b0047

src/activelearning/active_learning.py
pedromsferraz-mila: This is something I considered in the design, but I went with a budget-agnostic `Oracle` to keep the components decoupled.

If we add budget logic into the `Oracle` itself, it becomes harder to use them in isolation, like in notebooks, unit tests, or benchmarks where 'budgets' don't really exist. Right now, the active learning loop handles the management, and since `oracle.query` is only called after the loop verifies costs, there's no immediate risk of 'runaway' queries.

If we decide we need harder enforcement later, we could implement a `BudgetedOracle` wrapper. That way we get the safety without tying in all oracle implementations with budget logic.

src/activelearning/oracle/composite_oracle.py
pedromsferraz-mila: Good point, fixed! 65f30318120ac8f685f2c77ce1598d30aae993aa

src/activelearning/surrogate/dummy_mean_surrogate.py
alexhernandezgarcia: Good, thanks for explaining!

src/activelearning/sampler/pool_score_sampler.py
alexhernandezgarcia: Great, I agree. Thanks for clarifying the docstrings too!

src/activelearning/active_learning.py
alexhernandezgarcia: Ok, I agree. I don't think this will be a major issue, but I wanted to have your opinion. Completely fine to leave it as is for now.

alexhernandezgarcia (approved):
Looks good, ready to merge as far as I can tell! Thanks for the careful review of my comments!

alexhernandezgarcia:
Oh I forgot to say that I didn't review the tests in depth, as you have probably figured :sweat_smile: There's way too many for me to provide a meaningful review without delaying this too much. If that's ok with you, it's ok with me.

PR #3, opened 2026-02-26
feat: BoTorch surrogate with multi-fidelity, partial updates, and custom kernel support

olutayo-mila:
## Summary

Implements `BoTorchSurrogate`, a flexible Gaussian Process surrogate backed by BoTorch and GPyTorch.

### Core implementation
- **Single-fidelity and multi-fidelity GP routing** — automatically selects `SingleTaskGP` or `SingleTaskMultiFidelityGP` based on observation data
- **Fidelity confidence mapping** — integer fidelity IDs are mapped to continuous confidence values before being passed to the model
- **Full refit and partial (Cholesky) update** — controlled via `use_partial_updates` flag
- **Custom kernel support** — pass any GPyTorch `covar_module`; set `custom_fidelity_kernel=True` to bypass BoTorch's built-in MF model
- **Custom optimizer support** — pass a `custom_fit_function` (e.g. an Adam loop for Deep Kernel Learning) to replace BoTorch's default L-BFGS-B
- **State dict save/load** — hyperparameters can be extracted and injected, including pre-fit injection via a pending state dict

### Bug fixes
- `_is_multi_fidelity` flag is now reset on every `_parse_observations` call, preventing stale state when re-fitting on different data types
- `_partial_update` now guards against empty observation lists instead of crashing with `torch.stack([])`

### Tests (25 tests, all passing)
Covers: model routing, tensor shapes, error paths (`predict` before fit, missing fidelity on candidates), confidence mapping for both observations and candidates, full/partial update modes, state dict round-trip, custom kernel/optimizer hooks, transform flags, and all regression cases for the two bug fixes.

### Other
- Added `scripts/` to `.gitignore` (local experiment scripts)


src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,383 @@
...
+        self._train_X = torch.cat([self._train_X, new_X], dim=0)
+        self._train_Y = torch.cat([self._train_Y, new_Y], dim=0)
+
+        # Fast Cholesky update (internal data scaling transforms apply automatically)
+        # Note: condition_on_observations requires the model to have made at least one prediction
+        # to populate internal caches. We need to trigger a prediction first.
+        self.model.eval()
+        with torch.no_grad():
+            # Make a dummy prediction to initialize caches
+            _ = self.model.posterior(new_X[:1])
```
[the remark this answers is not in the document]
olutayo-mila: Not necessary as this is not a BoTorch version issue. Also the reason for the workaround is stated clearly in the comments.

tests/surrogate/test_botorch_surrogate.py
```diff
@@ -0,0 +1,464 @@
+import pytest
+import torch
+from botorch.models import SingleTaskGP
+from botorch.models.gp_regression_fidelity import SingleTaskMultiFidelityGP
+
+# Adjust import paths based on your actual src structure
```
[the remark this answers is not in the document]
olutayo-mila: Fixed.

src/activelearning/active_learning.py
```diff
@@ -63,8 +63,8 @@ def active_learning(
         # Note: get_observations_iterable() is called multiple times to avoid
         # consuming the same iterable across multiple consumers
 
-        # Fit surrogate on current dataset observations and update acquisition function
-        surrogate.fit(dataset.get_observations_iterable())
+        # Update surrogate with current dataset (decides internally whether to do full refit or partial update)
+        surrogate.update(dataset)
```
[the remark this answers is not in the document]
olutayo-mila: No need to worry about backward compatibility issues here as the library is just being built.

pyproject.toml
```diff
@@ -10,13 +10,15 @@ readme = "README.md"
...
     "numpy>=1.26.0,<3.0.0",
+    "botorch>=0.16.1",
     "torch>=2.5.1",
 ]
 
 [dependency-groups]
 dev = [
     "pytest>=7.0",
     "pre-commit>=3.0",
+    "matplotlib>=3.10.8",
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,391 @@
...
+        # 2. Build Model Configuration
+        # If MF is true AND they aren't using a custom fidelity kernel, use BoTorch's default MF model
+        if self._is_multi_fidelity and not self.custom_fidelity_kernel:
+            self.model = SingleTaskMultiFidelityGP(
+                self._train_X, 
+                self._train_Y, 
+                data_fidelities=[-1],
+                outcome_transform=outcome_transform,
+                input_transform=input_transform
+            )
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,391 @@
...
+            y_data.append(torch.as_tensor(obs.y, dtype=torch.float64))
+           
+            if obs.fidelity is not None:
+                # Map the integer ID to the continuous confidence value.
+                mapped_fid = self._fidelity_confidences.get(obs.fidelity, obs.fidelity)
+                fidelities.append(mapped_fid)
+
+        train_X = torch.stack(x_data)
+        train_Y = torch.stack(y_data).view(-1, 1)  # BoTorch expects shape (n, 1)
+
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,391 @@
...
+        obs_list = list(new_observations)
+        if not obs_list:
+            return
+
+        new_X, new_Y = self._parse_observations(obs_list)
+
+        # _train_X/_train_Y are non-None here: the guard at the top of this
+        # method returns early when self._train_X is None.
+        assert self._train_X is not None and self._train_Y is not None
+
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed.

tests/surrogate/test_botorch_surrogate.py
```diff
@@ -0,0 +1,462 @@
...
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,383 @@
...
+from botorch.models.gp_regression_fidelity import SingleTaskMultiFidelityGP
+from botorch.fit import fit_gpytorch_mll
+from botorch.models.transforms.outcome import Standardize
+from botorch.models.transforms.input import Normalize
+from gpytorch.mlls import ExactMarginalLogLikelihood
+from gpytorch.module import Module
+
+from activelearning.surrogate.surrogate import Surrogate
+from activelearning.utils.types import Candidate, Observation
+from activelearning.dataset.dataset import Dataset
```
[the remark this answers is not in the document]
olutayo-mila: There was no need for TYPE_CHECKING in the DummySurrogate in the first place. Fixed.

src/activelearning/dataset/dataset.py
```diff
@@ -63,6 +63,47 @@ def get_observations_iterable(self):
...
+            Returns an empty iterable if no observations have been added yet.
+
+        Notes
+        -----
+            - This method should return only the observations from the last
+              add_observations() call, not all historical observations
+            - Like get_observations_iterable(), this may be called multiple times
+              and should return a fresh iterable each time
+            - If add_observations() is called again, subsequent calls to this method
+              should return the new batch, not the previous one
```
alexhernandezgarcia: Reduced indedation one level.
```suggestion
        - This method should return only the observations from the last
          add_observations() call, not all historical observations
        - Like get_observations_iterable(), this may be called multiple times
          and should return a fresh iterable each time
        - If add_observations() is called again, subsequent calls to this method
          should return the new batch, not the previous one
```

src/activelearning/dataset/dataset.py
```diff
@@ -63,6 +63,47 @@ def get_observations_iterable(self):
...
+            latest = list(dataset.get_latest_observations_iterable())
+            # latest == [obs1, obs2, obs3]
+            ```
+
+            After second batch:
+            ```python
+            dataset.add_observations([obs4, obs5])
+            latest = list(dataset.get_latest_observations_iterable())
+            # latest == [obs4, obs5]  (not obs1-obs5)
+            ```
```
alexhernandezgarcia: Reduce indentation one level.
```suggestion
        After first batch:
        ```python
        dataset.add_observations([obs1, obs2, obs3])
        latest = list(dataset.get_latest_observations_iterable())
        # latest == [obs1, obs2, obs3]
        ```

        After second batch:
        ```python
        dataset.add_observations([obs4, obs5])
        latest = list(dataset.get_latest_observations_iterable())
        # latest == [obs4, obs5]  (not obs1-obs5)
        ```
```

src/activelearning/dataset/dataset.py
```diff
@@ -55,6 +55,47 @@ def get_observations_iterable(self):
...
+        -------
+        latest_observations_iterable : Iterable[Observation]
+            Iterable of observations from the most recent add_observations() call.
+            Returns an empty iterable if no observations have been added yet.
+
+        Notes
+        -----
+            - This method should return only the observations from the last
+              add_observations() call, not all historical observations
+            - Like get_observations_iterable(), this may be called multiple times
```
pedromsferraz-mila: Given the latest changes (merged from main), we should mention that calling this multiple times should provide an iterable for the same sequence of observations, in the same order. We should also mention that implementations are responsible for pinning any randomness.

src/activelearning/surrogate/surrogate.py
```diff
@@ -18,28 +26,70 @@ class Surrogate(ABC):
...
+            observations via ``update()``, ``False`` if it should be fully
+            retrained via ``fit()`` on all observations.
+        """
+        pass
+
+    @abstractmethod
+    def update(self, observations: Iterable[Observation]) -> None:
+        """Incrementally update the surrogate with the latest (new) observations.
+
+        Called by the active learning loop only when ``updates_from_latest()``
```
pedromsferraz-mila: The current comment on update seems overly specific, focusing on incremental updates like low-rank Cholesky conditioning. Since this method could also be applicable in broader contexts, such as online or continuous learning, consider rephrasing to make it more general and inclusive of other potential use cases.

src/activelearning/surrogate/surrogate.py
```diff
@@ -18,28 +26,70 @@ class Surrogate(ABC):
...
+        Returns
+        -------
+        bool
+            ``True`` if this surrogate should receive only the latest (new)
+            observations via ``update()``, ``False`` if it should be fully
+            retrained via ``fit()`` on all observations.
+        """
+        pass
+
+    @abstractmethod
```
pedromsferraz-mila: I noticed that `update` is defined as an `@abstractmethod` while `fit` uses `NotImplementedError`. This creates an inconsistent requirement for subclasses. If the goal is to let users choose which one to implement based on their data flow, we should probably use `NotImplementedError` for both to keep the interface flexible.

src/activelearning/surrogate/surrogate.py
```diff
@@ -18,28 +26,70 @@ class Surrogate(ABC):
...
+        ------
+        NotImplementedError
+            If this surrogate does not implement fit().
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not implement fit(). "
+            "Override fit() or ensure updates_from_latest() returns True and override update()."
+        )
+
     def set_fidelity_confidences(self, confidences: dict[int, float]) -> None:
```
pedromsferraz-mila, line 124: Since `set_fidelity_confidences` is invoked within the active learning loop for all surrogates, we should make it an `@abstractmethod`. This ensures that any new subclass provides an implementation.

src/activelearning/surrogate/surrogate.py
```diff
@@ -10,6 +10,14 @@ class Surrogate(ABC):
     Surrogate models approximate the oracles based on
     observed data, enabling efficient candidate evaluation.
 
+    The active learning loop calls ``updates_from_latest()`` each round to decide
```
pedromsferraz-mila: Could we rephrase this from the surrogate's perspective, focusing less on the active learning loop? For example: "Surrogate must implement `updates_from_latest` to determine whether to update using only the latest observations (for incremental updates) or all observations (for full retraining).

src/activelearning/surrogate/surrogate.py
```diff
@@ -18,28 +26,70 @@ class Surrogate(ABC):
...
 
     @abstractmethod
-    def fit(self, observations: Iterable[Observation]) -> None:
-        """Fit the surrogate model to observations.
+    def updates_from_latest(self) -> bool:
+        """Declare whether this surrogate updates incrementally from latest observations.
+
+        Called by the active learning loop each round to determine which iterable
+        to provide:
+          - Return ``True``  → loop calls ``update(dataset.get_latest_observations_iterable())``.
```
pedromsferraz-mila: Similarly here, consider revising the comments in the to focus on the surrogate's perspective, describing its behavior and requirements without referencing the active learning loop in detail. This will make the documentation more self-contained and easier to understand for users unfamiliar with the rest of the codebase.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,432 @@
...
+            Adam training loop for Deep Kernel Learning). If provided, this overrides
+            BoTorch's default L-BFGS-B optimizer.
+        covar_module : Module, optional
+            Optional custom GPyTorch kernel for non-fidelity dimensions. If None,
+            defaults to BoTorch's standard Matérn/RBF kernels.
+        use_partial_updates : bool, default=False
+            If True, update() uses fast incremental Cholesky updates when the model is
+            already fitted. If False, update() always performs full retraining for
+            maximum reliability. Beginners should use False.
+        custom_fidelity_kernel : bool, default=False
```
pedromsferraz-mila: If the `custom_fidelity_kernel` parameter is solely used to determine whether a custom `covar_module` is provided, it might be redundant. You could simplify the code by removing it from the `__init__` parameters and setting `self.custom_fidelity_kernel = covar_module is not None` directly in the constructor.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,432 @@
...
+            If True, bypasses BoTorch's default multi-fidelity model and allows the
+            custom `covar_module` to handle fidelity dimensions internally.
+        """
+        self.model = None
```
pedromsferraz-mila: Consider moving these down to the "Internal state tracking" block.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,432 @@
...
+        standardize_outputs: bool = True,
+        optimize_hyperparameters: bool = True,
+        fit_kwargs: Optional[dict[str, Any]] = None,
+        custom_fit_function: Optional[Callable] = None,
+        covar_module: Optional[Module] = None,
+        use_partial_updates: bool = False,
+        custom_fidelity_kernel: bool = False,
+    ):
+        """
+        Parameters
```
pedromsferraz-mila, line 40: If seems like some of these parameters might interact with each other. If that's the case, consider adding assertions or checks to validate parameter combinations. This will help catch inconsistent configurations early and provide informative error messages to guide users. For example, you could check for conflicts between `custom_fit_function` and `optimize_hyperparameters`.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,432 @@
...
+            contain a fidelity value.
+        """
+        # Reset on every call so that switching from MF → SF (or vice versa)
+        # never leaves stale state from a previous invocation.
+        self._is_multi_fidelity = False
+
+        x_data, y_data, fidelities = [], [], []
+        obs_list = list(observations)
+
+        for obs in obs_list:
```
pedromsferraz-mila: This implementation is general and works well with any `Iterable`, which is a strength. However, iterating over all elements in the loop could be a performance bottleneck for large datasets. To improve efficiency, consider creating a helper method in `utils/types.py` that converts `Iterable[Observation]` into a tensor. This method could first check if the `Iterable` is a `Sequence` and use faster batch conversions in that case, while falling back to the current `for` loop for other types of `Iterable`. This would optimize performance for common use cases without sacrificing generality.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,432 @@
...
+
+        Raises
+        ------
+        ValueError
+            If the model was trained on multi-fidelity data, but candidates are
+            missing fidelity values.
+        """
+        x_data, fidelities = [], []
+
+        for cand in candidates:
```
pedromsferraz-mila: Same as in `_parse_observations`, this could be abstracted away into a helper method under `utils/types.py`.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,432 @@
...
+            fid_tensor = torch.tensor(fidelities, dtype=torch.float64).view(-1, 1)
+            test_X = torch.cat([test_X, fid_tensor], dim=-1)
+
+        return test_X
+
+    def _build_and_fit_model(self) -> None:
+        """Constructs and optionally optimizes the Gaussian Process model."""
+        # _build_and_fit_model is only called after _parse_observations has populated
+        # these tensors; assert here so the type checker sees them as non-None.
+        assert self._train_X is not None and self._train_Y is not None
```
pedromsferraz-mila: Add an assertion message to clarify that `_parse_observations` must be called before invoking this method.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,432 @@
...
+        ----------
+        new_observations : Iterable[Observation]
+            Iterable of strictly new observations to condition the GP on.
+        """
+        if self.model is None or self._train_X is None:
+            # Fallback to a full fit if the model hasn't been initialized
+            self.fit(new_observations)
+            return
+
+        obs_list = list(new_observations)
```
pedromsferraz-mila: It's unclear whether `new_observations` can always be safely converted to a list (e.g., it might be a `DataLoader` or another iterable that does not support direct conversion). A safer alternative would be to iterate over `new_observations` directly without converting it to a list. This avoids potential issues with non-list iterables.

src/activelearning/surrogate/dummy_mean_surrogate.py
```diff
@@ -70,3 +70,30 @@ def predict(self, candidates: Sequence[Candidate]) -> Mapping[str, list[float]]:
...
+        """Return False — DummyMeanSurrogate always performs full retraining.
+
+        Returns
+        -------
+        bool
+            Always ``False``.
+        """
+        return False
+
+    def update(self, observations: Iterable[Observation]) -> None:
```
pedromsferraz-mila: Since `DummyMeanSurrogate` always returns `False` from `updates_from_latest`, the `update` method is unnecessary here.

pedromsferraz-mila (commented):
Nice work! Left a few small notes throughout, but overall it's looking very solid.

src/activelearning/surrogate/dummy_mean_surrogate.py
olutayo-mila: Good point. Fixed.

src/activelearning/surrogate/surrogate.py
olutayo-mila: Good point. This was fixed in a previous commit!

src/activelearning/surrogate/surrogate.py
olutayo-mila: Fair point. However, there could be instances wherein a surrogate doesn't require fidelity confidences or the AL setting is single fidelity. So, I think it might be better to leave it as is. Let me know what you think.

pedromsferraz-mila: Hm, that's a good point, but in that case we should change the call in the active learning loop. Currently, on L60 of `src/activelearning/active_learning.py`, this method of the surrogate is always called: `surrogate.set_fidelity_confidences(oracle.get_fidelity_confidences())`

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Thanks for the suggestion! custom_fidelity_kernel and covar_module serve different purposes:

- In single-fidelity SingleTaskGP, covar_module is optional (defaults are used if None).

- In multi-fidelity, setting custom_fidelity_kernel=True means we bypass SingleTaskMultiFidelityGP and expect the provided covar_module to model fidelity internally.

Therefore, custom_fidelity_kernel cannot be inferred from covar_module is not None: a user may provide covar_module in single-fidelity without intending any fidelity handling.

I updated the docstrings to make this clearer(commit b12c878b2874b82cf456256bb75ce10ea1f4a99b)

src/activelearning/surrogate/surrogate.py
pedromsferraz-mila: Good point, let's keep as-is.

olutayo-mila: I agree. Also I added a comment just to make it clearer (76233f59fdbdcc951f324ebedcb36387fe8dd156).

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Great point, thanks! I have address the issue by adding observations_to_tensors(), and candidates_to_tensor() to utils/types.py and updated botorch_surrogate.py accordingly. See here (8f19c69d372f15477d2b0b2050f8eca34d9e76f9)

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: See above comment. Thanks.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point. Fixed (6c0651dbfe97886ec2c245f1060e79d43eb5d6a0). Note that I made some other changes in this commit, sorry about the clutter. Specifically, I edited the new functions I added (observations_to_tensors and candidates_to_tensors) to be cleaner. Due to this change, I updated the multi-fidelity tests in test_botorch_surrogate.py to call the set_fidelity_confidences method using hardcoded dicts.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Thanks for pointing this out, that’s a good consideration. Iterating directly over the iterable would work if we only needed a single pass. In this method, however, we need to (1) check for emptiness, (2) validate fidelities (and report indices if missing), and (3) construct the tensors for conditioning the GP.

We could combine (1) and (2) into a single streaming loop, but we would still need to buffer the observations in memory in order to build the tensors afterward. So even in a single-pass implementation, we would effectively be materializing the batch anyway. Using list(new_observations) keeps that intent explicit and the logic simpler.

In practice, new_observations here represents the latest oracle batch in the active learning loop, so it is expected to be a small, finite set. To make that contract clearer, I’ve updated the docstring to explicitly state that this iterable is assumed to be a finite batch that will be materialized internally. See here (6eb57d39b2326861bfd42b1f0acf10a243849418)

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Agree! Fixed (5ea1473ba88d63d7345308dcd12c5abbda69b203)

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Following our discussion, I have made the routing logic clearer so that model architecture is now determined by whether a covar_module is provided and the _is_multi_fidelity flag. I removed the custom_fidelity_kernel flag. See here: 9ca7e9656209844f1a2df8554afcdab65db710e1

pedromsferraz-mila: That looks great, thanks!

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Thanks, the only variables that interact with each other now are custom_fit_function and optimize_hyperparameters, and I have now added a check. See here 9ca7e9656209844f1a2df8554afcdab65db710e1.

pedromsferraz-mila: Perfect, thanks!

src/activelearning/dataset/dataset.py
olutayo-mila: Done. See here e82ca5576ff36a67ee73c814e4d215f316263e1d.

pedromsferraz-mila: Perfect, thanks!

src/activelearning/surrogate/surrogate.py
olutayo-mila: Done. See here ce7f27a245d1c3c995059acc5a4b6731fb53f785.

src/activelearning/surrogate/surrogate.py
olutayo-mila: Done. See here ce7f27a245d1c3c995059acc5a4b6731fb53f785.

src/activelearning/surrogate/surrogate.py
olutayo-mila: Done. See here ce7f27a245d1c3c995059acc5a4b6731fb53f785.

src/activelearning/surrogate/surrogate.py
pedromsferraz-mila: That looks great, thanks!

src/activelearning/surrogate/surrogate.py
pedromsferraz-mila: Perfect, thanks!

src/activelearning/surrogate/surrogate.py
pedromsferraz-mila: Great, thanks!

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Following our conversation, I edited the functions observations_to_tensors() and candidates_to_tensor() in utils/types.py to be more flexible. Instead of forcing the shape of the tensor to 2d (BoTorch Style), we keep the same shape as the observations and the user can then reshape as needed. I think it is cleaner this way. I also renamed "train_X" and "train_y" to just X and y. Let me know what you think. See here 81796cfc26e591248e8a9370742db67f7c652e37.

pedromsferraz-mila: That looks good to me, thanks!

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Following our conversation, I updated the docstrings to make it clear that Iterable will be materialized to list. See here 55cddaf807a7d28ef69e92ba089fad7d0d2f1d48. I also changed Sequences to Iterables in this commit. Let me know your thoughts. Thanks.

pedromsferraz-mila: Perfect, thank you!

src/activelearning/dataset/list_dataset.py
```diff
@@ -20,6 +21,7 @@ def add_observations(self, observations: Sequence[Observation]) -> None:
             Sequence of observations to add.
         """
         self._records.extend(observations)
+        self._latest_records = list(observations)
```
alexhernandezgarcia: Suggestion: would it not be better to store the _indices_ of the latest records (in `self._records`) and have `get_latest_observations_iterable()` retrieve the observations at these indices in `self._records`? This would avoid duplicating data storage,

src/activelearning/dataset/list_dataset.py
```diff
@@ -31,6 +33,17 @@ def get_observations_iterable(self) -> list[Observation]:
...
+    def get_latest_observations_iterable(self) -> list[Observation]:
+        """Retrieve the most recently added observations.
+
+        Returns
+        -------
+        latest_observations_iterable : list[Observation]
+            List of observations from the most recent add_observations() call.
+            Returns an empty list if no observations have been added yet.
+        """
+        return list(self._latest_records)
```
alexhernandezgarcia: Note that `list(...)` is used both here and in `self._latest_records = list(observations)`. Is it intentional?

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+from activelearning.surrogate.surrogate import Surrogate
+from activelearning.utils.types import (
+    Candidate,
+    Observation,
+    observations_to_tensors,
+    candidates_to_tensor,
+)
+
+
+class BoTorchSurrogate(Surrogate):
```
alexhernandezgarcia: Suggestion: add "GP" to the class name, for example `GPBotorchSurrogate` or `GPSurrogateBotorch`

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+        optimize_hyperparameters: bool = True,
+        fit_kwargs: Optional[dict[str, Any]] = None,
+        custom_fit_function: Optional[Callable] = None,
+        covar_module: Optional[Module] = None,
+        use_partial_updates: bool = False,
+    ):
+        """
+        Parameters
+        ----------
+        normalize_inputs : bool, default=True
```
alexhernandezgarcia: If the scaling is into `[0, 1]`, I would call the flag `scale_inputs` since "normalize" may be interpreted as "subtract mean and divide by standard deviation".

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+            If True, scales X values to [0, 1] internally.
+        standardize_outputs : bool, default=True
+            If True, scales Y values to mean=0, var=1 internally.
```
alexhernandezgarcia:
```suggestion
            If True, normalizes Y values to mean=0, var=1 internally.
```

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+        # Toggles and configurations
+        self.normalize_inputs = normalize_inputs
+        self.standardize_outputs = standardize_outputs
+        self.optimize_hyperparameters = optimize_hyperparameters
+        self.fit_kwargs = fit_kwargs or {}
+        self.custom_fit_function = custom_fit_function
+        self.covar_module = covar_module
+        self.use_partial_updates = use_partial_updates
+
+        if custom_fit_function is not None and not optimize_hyperparameters:
```
alexhernandezgarcia: Very minor, but I guess this check could precede the "Toggles and configurations" part to fail earlier. It actually does not use the `self.___` attributes, I guess intentionally?

src/activelearning/utils/types.py
```diff
@@ -45,26 +47,141 @@ class Observation:
...
+        Input features tensor. Shape is determined by the input data.
+    y : torch.Tensor
+        Output labels tensor. Shape is determined by the input data.
+    fidelities : list[float]
+        Mapped fidelity confidence values; empty if single-fidelity.
+    """
+    obs_list = observations if isinstance(observations, list) else list(observations)
+
+    X = _to_tensor([obs.x for obs in obs_list], torch.float64)
+    y = _to_tensor([obs.y for obs in obs_list], torch.float64)
```
alexhernandezgarcia: Similar to [my comment in another PR](https://github.com/milaforscience/activelearning/pull/4/changes#r2902514401): are we going to enforce using 64-bits floats everywhere? (happy to have this conversation outside this PR)

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+            as the final column if multi-fidelity.
+
+        Raises
+        ------
+        ValueError
+            If the model was trained on multi-fidelity data, but candidates are
+            missing fidelity values.
+        """
+        cand_list = candidates if isinstance(candidates, list) else list(candidates)
+        test_X, fidelities = candidates_to_tensor(cand_list, self._fidelity_confidences)
```
alexhernandezgarcia: This would create three copies of the candidates: `candidates`, `cand_list` and `test_X`. Do we need all three?

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+                raise ValueError(
+                    "Surrogate was fitted on multi-fidelity data. Candidates require fidelities."
+                )
+
+            fid_tensor = torch.tensor(fidelities, dtype=torch.float64).view(-1, 1)
+            test_X = torch.cat([test_X, fid_tensor], dim=-1)
+
+        return test_X
+
+    def _build_and_fit_model(self) -> None:
```
alexhernandezgarcia: I find it a bit confusing that there is a method `_build_and_fit_model()` and a `fit()` method that calls the former. What was the rationale? Could it not all be in `fit()`? Or perhaps the operations in  `_build_and_fit_model()` could be split into a `_build_model()` and `fit()`?

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+        """Constructs and optionally optimizes the Gaussian Process model."""
+        assert self._train_X is not None and self._train_Y is not None, (
+            "_parse_observations() must be called before _build_and_fit_model()."
+        )
```
alexhernandezgarcia: I believe `_parse_observations()` would not set `self._train_X` and `self._train_Y`, so the assertion message may be confusing.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+        if self._is_multi_fidelity and self.covar_module is None:
+            self.model = SingleTaskMultiFidelityGP(
+                self._train_X,
+                self._train_Y,
+                data_fidelities=[-1],
+                outcome_transform=outcome_transform,
+                input_transform=input_transform,
+            )
+        else:
+            # Single-fidelity OR multi-fidelity with a custom kernel
```
alexhernandezgarcia, line 558: Maybe add a comment saying that if the "custom" kernel is None, it defaults to an RBF kernel (unless [BoTorch](https://botorch.readthedocs.io/en/stable/_modules/botorch/models/gp_regression.html) changes it)

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+        # 4. Optimize Hyperparameters (if toggled on)
+        if self.optimize_hyperparameters:
+            if self.custom_fit_function is not None:
+                # Use the user's custom fit function if provided
+                self.custom_fit_function(self.mll, **self.fit_kwargs)
+            else:
+                # Default BoTorch fitting procedure
+                fit_gpytorch_mll(self.mll, **self.fit_kwargs)
+
+        self.model.eval()
```
alexhernandezgarcia, line 231: What does `model.eval()` do exactly? Could you please add a few words about it as a comment?

src/activelearning/surrogate/botorch_surrogate.py
alexhernandezgarcia: Actually, I'm even wondering if `train_X` and `train_Y` need to be set as attributes. They could also be arguments of this method, no?

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+            If called before the surrogate model has been fitted.
+        """
+        if self.model is None:
+            raise RuntimeError(
+                "Surrogate model must be fitted before calling predict()."
+            )
+
+        test_X = self._parse_candidates(candidates)
+
+        self.model.eval()
```
alexhernandezgarcia: Does `eval()` need to be called here (again)?

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,430 @@
...
+        Called by the active learning loop only when ``updates_from_latest()``
+        returns ``True``. Uses a fast low-rank Cholesky conditioning step without
+        retraining hyperparameters.
+
+        Parameters
+        ----------
+        observations : Iterable[Observation]
+            Iterable of the most recent observations to condition the GP on.
+        """
+        self._partial_update(observations)
```
alexhernandezgarcia: Why do we need both `update()` and `_partial_update()`?

src/activelearning/utils/types.py
```diff
@@ -45,26 +47,141 @@ class Observation:
...
     -------
-    result : Sequence[Observation]
-        Sequence of Observation objects, where each observation combines the
+    result : list[Observation]
+        List of Observation objects, where each observation combines the
         candidate's x and fidelity with its label as y.
     """
-    if len(candidates) != len(labels):
+    candidates_list = candidates if isinstance(candidates, list) else list(candidates)
+    labels_list = labels if isinstance(labels, list) else list(labels)
```
alexhernandezgarcia, line 68: Why is it needed to materialise `candidates` and `labels` as lists / why the previous version that iterates over the inputs isn't good?

src/activelearning/utils/types.py
```diff
@@ -45,26 +47,141 @@ class Observation:
...
+    Returns
+    -------
+    X : torch.Tensor
+        Input features tensor. Shape is determined by the input data.
+    y : torch.Tensor
+        Output labels tensor. Shape is determined by the input data.
+    fidelities : list[float]
+        Mapped fidelity confidence values; empty if single-fidelity.
+    """
+    obs_list = observations if isinstance(observations, list) else list(observations)
```
alexhernandezgarcia: Same question as before: why is conversion to list needed?

src/activelearning/utils/types.py
```diff
@@ -45,26 +47,141 @@ class Observation:
...
+    y = _to_tensor([obs.y for obs in obs_list], torch.float64)
+    fidelities = (
+        [
+            fidelity_confidences[obs.fidelity]
+            for obs in obs_list
+            if obs.fidelity is not None
+        ]
+        if fidelity_confidences
+        else []
+    )
```
alexhernandezgarcia: This is a bit hard to parse as one-liner, no?

alexhernandezgarcia (commented):
Very nice work! I mostly have questions on the implementation of the GP surrogate. I found a few things slightly confusing. Hopefully we'll be able to quickly discuss them in the comments.

src/activelearning/dataset/list_dataset.py
olutayo-mila: Yes, it will be more efficient this way. Thanks for the suggestion. Fixed 6e15e61fc9b16f6cc09914e2fb9d98d12fdb5594.

src/activelearning/dataset/list_dataset.py
olutayo-mila: Yes, it is intentional. It is a defensive copy so that any external mutation of the returned list does not affect the internal state.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: I agree. Fixed 83891947788b8d44903ebee5315986196cca01f6.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: I agree. Fixed 6e5dbc0d9fdaedf8f41fa97aebf2e5caf3386f87.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Thanks. Fixed 6e5dbc0d9fdaedf8f41fa97aebf2e5caf3386f87.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Yes it was intentional. However, I agree that the check can be moved earlier to fail fast. Fixed 2175f7339a5fb3611334bdb947768ac3a4ade16e.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point! Fixed 7cf08995a4207fdd59fa98c4ec8fa475b53ad595.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point, thanks for flagging this. The original intent was to separate concerns internally, but I agree the _build_and_fit_model() + fit() combination was confusing. I refactored the code following your suggestion: _build_model() now handles construction, and fit() is responsible for the fitting logic. This should make the responsibilities clearer. See here: 2f816458a63b9f8e6e141da5558c4f2cc672bae0.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good catch, thanks. You are correct that _parse_observations() does not set self._train_X / self._train_Y, so the assertion message was confusing. I refactored this so _build_model(train_X, train_Y) now takes the tensors explicitly and is called from fit() after parsing the observations. This removes the implicit state dependency and the assertion. I kept _train_X and _train_Y as internal attributes since they are still needed for _partial_update() and historical state tracking, but _build_model() no longer relies on them. See here 08148261b77b4e6853b5f9caa0c845186d47d898.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point. Done e9dd1ecf9603070869376987e1065f9823fc9f04.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point, thanks. model.eval() switches the PyTorch module to evaluation mode so the model is used in inference mode rather than training mode. This is also the standard pattern used in BoTorch tutorials before computing posteriors or evaluating acquisition functions. I added a short clarifying comment in the code. See here 599052a67a86ce1ca267d4b2bf3adc828e9b5588.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: It’s not strictly required. I kept it here as a defensive safeguard to ensure the model is in inference mode in case the mode was changed between fit() and predict().

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point, thanks. The helper came from an earlier version of the design, but there’s no real need to keep both methods. I’ve simplified this by moving the logic into `update()` and removing `_partial_update()`. See here b6f60af7c1d58961e3e2ce839d840b70227e5e63.

src/activelearning/utils/types.py
olutayo-mila: Good point. The materialization is intentional because the inputs are now typed as Iterable, so they may be one-pass generators rather than reusable containers. Converting to a list avoids iterator exhaustion and makes repeated access deterministic. Specifically, in label_candidates(), this lets us validate matching lengths before building the observations. In observations_to_tensors() and candidates_to_tensor(), we also need multiple passes over the same data to extract different fields.

src/activelearning/utils/types.py
olutayo-mila: I agree. Fixed 9a875d6c825a3913458c8173f4d55cc78a534195.

src/activelearning/dataset/dataset.py
olutayo-mila: Done.

src/activelearning/dataset/dataset.py
olutayo-mila: Done.

alexhernandezgarcia (commented):
Submitting part-way (incomplete) review

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,707 @@
...
+    def __init__(
+        self,
+        scale_inputs: bool = True,
+        standardize_outputs: bool = True,
+        optimize_hyperparameters: bool = True,
+        fit_kwargs: Optional[dict[str, Any]] = None,
+        custom_fit_function: Optional[Callable[..., Any]] = None,
+        covar_module: Optional[Module] = None,
+        use_partial_updates: bool = False,
+    ) -> None:
```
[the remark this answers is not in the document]
olutayo-mila: @copilot apply changes based on [this feedback](https://github.com/milaforscience/activelearning/pull/3#discussion_r2928053716)

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,707 @@
...
+            Model parameters to load.
+        """
+        if self.model is not None:
+            self.model.load_state_dict(state_dict)
+        else:
+            # Store it for when _build_model is called
+            self._pending_state_dict = state_dict
+
+    # ------------------------------------------------------------------
+    # BoTorch-specific public interface for acquisition modules
```
pedromsferraz-mila: Since these methods are only intended for the acquisition function and aren’t called in this PR, would it make sense to move them to the acquisition function PR? It’ll be easier to review them with the full context so we can decide if they belong here, in the acquisition class, or as standalone helpers.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,707 @@
...
+        observations : Iterable[Observation]
+            Iterable of observations to train on. Materialized to a list internally.
+
+        Raises
+        ------
+        ValueError
+            If observations is empty or structurally incompatible.
+        """
+        obs_list = list(observations)
+        if not obs_list:
```
pedromsferraz-mila, line 143: The BoTorch surrogate needs to handle empty datasets, since the AL loop starts with empty observations. We could use an uninformative prior, or let the `fit` method be a no-op and make the acquisition function uniform scores if the surrogate hasn't been fit yet.

Could we also update the fixture in `test_active_learning.py` to use the new surrogate? That'll give us a good end-to-end test for the whole AL loop.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Agreed. I have moved most of the methods relevant to the acquisition module to the acquisition branch. I also edited the relevant tests accordingly. See here 024cbde5d2c62a2cfeee47c46680f3876207f627.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point, thanks! I have updated the BoTorch surrogate so that in the scenario of empty datasets, the fit() method defaults to no-op. I also added an is_fitted method to the surrogate which is used in the AL-loop to decide whether to update the acquisition or use default (e.g. return a constant score for all candidates). Moreover, I updated the relevant tests. See here 0b86aff66132f11343e8df0af529757748216e31. 

Ps: I made a slight edits to the doc string for the is_fitted method in surrogate.py here 818052d99917b815861ac9d493fb99f21e5b9e59 and here 4869ae0932fe03a8df0306d6cd285208f4d7154c.

src/activelearning/surrogate/surrogate.py
```diff
@@ -18,27 +22,104 @@ class Surrogate(ABC):
...
-            for large datasets.
+            Iterable of the most recent observations (from the latest oracle query).
 
         Notes
         -----
-        Implementations should validate their input requirements:
-        - If you need len() or multiple passes, materialize to list or assert Sequence
-        - If you can handle streaming data, consume the iterable directly
+            This method is *not* called when ``updates_from_latest()`` returns
+            ``False``. In that case the loop calls ``fit(observations)`` instead.
```
alexhernandezgarcia: Minor wording suggestion:
```suggestion
            This method is *never* called if ``updates_from_latest()`` returns
            ``False``. In that case the loop calls ``fit(observations)`` instead.
```

src/activelearning/surrogate/surrogate.py
```diff
@@ -18,27 +22,104 @@ class Surrogate(ABC):
...
-        pass
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not implement update(). "
+            "Override update() or ensure updates_from_latest() returns False and override fit()."
+        )
+
+    def fit(self, observations: Iterable[Observation]) -> None:
+        """Fit the surrogate model to observations from scratch.
+
+        Called by the active learning loop when ``updates_from_latest()`` returns
```
alexhernandezgarcia: Minor wording suggestion:
```suggestion
        Called by the active learning loop if ``updates_from_latest()`` returns
```

src/activelearning/utils/types.py
```diff
@@ -45,26 +47,157 @@ class Observation:
 
 
 def label_candidates(
-    candidates: Sequence[Candidate], labels: Sequence[Any]
-) -> Sequence[Observation]:
-    """Convert a sequence of candidates and their corresponding labels into observations.
+    candidates: Iterable[Candidate], labels: Iterable[Any]
```
alexhernandezgarcia, line 50: Can you say a few words about the change from `Sequence` to `Iterable`?

src/activelearning/utils/types.py
```diff
@@ -45,26 +47,157 @@ class Observation:
...
+    ----------
+    candidates : Iterable[Candidate]
+        Candidates to convert. Materialized to a list internally.
+    fidelity_confidences : dict[int, float], optional
+        Mapping from integer fidelity IDs to continuous confidence values.
+        If None, candidates must not include fidelity values. If a mapping is
+        provided, every non-None fidelity must be present in it. Raises
+        ``ValueError`` when fidelities are present but no mapping is supplied,
+        and ``KeyError`` if a fidelity ID in the candidates is missing from the
+        mapping.
```
alexhernandezgarcia: Is it intentional to describe the raised errors here instead of under 
```
Raises
------
```
?

src/activelearning/utils/types.py
```diff
@@ -45,26 +47,157 @@ class Observation:
...
+    y : torch.Tensor
+        Output labels tensor. Shape is determined by the input data.
+    fidelities : list[float]
+        Mapped fidelity confidence values; empty if single-fidelity.
+    """
+    obs_list = observations if isinstance(observations, list) else list(observations)
+
+    X = _to_tensor([obs.x for obs in obs_list], torch.float64)
+    y = _to_tensor([obs.y for obs in obs_list], torch.float64)
+    has_fidelity = [obs.fidelity is not None for obs in obs_list]
```
alexhernandezgarcia: I am wondering whether it would not be better to create `X`, `y` and `has_fidelity` within  one single iteration over `obs_list` (or even over `obs` directly). I guess you have thought of this and decided for this option instead, would you mind explaining the decision?

src/activelearning/utils/types.py
```diff
@@ -45,26 +47,157 @@ class Observation:
...
+        Tensor with shape determined by the input data. Scalar inputs produce
+        shape (n,); array inputs produce shape (n, d).
+    """
+    try:
+        return torch.as_tensor(values, dtype=dtype)
+    except (TypeError, ValueError, RuntimeError):
+        return torch.stack([torch.as_tensor(v, dtype=dtype) for v in values])
+
+
+def observations_to_tensors(
```
alexhernandezgarcia, line 102: Just noticing that with the current implementation, it would kind of be possible to get `X` and the fidelities from `candidates_to_tensor()`. Do you think it would be worth it, or better to keep the methods independent even if they share most of the code?

src/activelearning/surrogate/botorch_surrogate.py
alexhernandezgarcia: Thanks!

src/activelearning/dataset/dataset.py
```diff
@@ -55,6 +55,51 @@ def get_observations_iterable(self):
...
+        latest = list(dataset.get_latest_observations_iterable())
+        # latest == [obs1, obs2, obs3]
+        ```
+
+        After second batch:
+        ```python
+        dataset.add_observations([obs4, obs5])
+        latest = list(dataset.get_latest_observations_iterable())
+        # latest == [obs4, obs5]  (not obs1-obs5)
+        ```
```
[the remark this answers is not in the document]
alexhernandezgarcia: If this is done, I think double colons (`::`) should be used to start literal blocks. Ref:
- https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html
- https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

src/activelearning/surrogate/surrogate.py
olutayo-mila: Agreed. Fixed.

src/activelearning/surrogate/surrogate.py
olutayo-mila: Agreed. Fixed.

src/activelearning/utils/types.py
olutayo-mila: Good point. I think there is some value in keeping the methods independent for readability, even if they share some logic. That said, I’m open to consolidating them if you think that's better.

src/activelearning/utils/types.py
olutayo-mila: Thanks, good point! I initially structured it that way for readability and ease of implementation. In hindsight, I agree that a single pass is cleaner and more efficient. I’ve updated both observations_to_tensors and candidates_to_tensor accordingly. See here 969e006dd2825f64394cc54f716cdb45ec1e0faf.

src/activelearning/utils/types.py
olutayo-mila: Good point. Fixed!

src/activelearning/utils/types.py
olutayo-mila: From my discussions with Pedro, the motivation for switching from Sequence to Iterable was to support a broader class of inputs. For example, PyTorch dataloaders are iterable but don’t implement the full Sequence interface.

src/activelearning/dataset/dataset.py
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed 0d7de66407a8a6105d2ddb76e8afa3739b656805.

src/activelearning/utils/types.py
alexhernandezgarcia: That makes sense, thanks!

src/activelearning/utils/types.py
alexhernandezgarcia: Thanks for the explanation 👍

src/activelearning/utils/types.py
alexhernandezgarcia: I didn't (and don't) have a strong opinion about this. I think the current version is fine, but I wanted to get your opinion :)

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,605 @@
...
+            # No-op on empty data: surrogate stays unfitted. The AL loop checks
+            # is_fitted() and will use random candidate selection for this round.
+            return
+
+        train_X, train_Y, is_multi_fidelity = self._parse_observations(obs_list)
+        self._is_multi_fidelity = is_multi_fidelity
+        self._train_X = train_X
+        self._train_Y = train_Y
+
+        self._build_model(train_X, train_Y)
```
alexhernandezgarcia: Here we could have only variables `self._train_X`, `self._train_Y` and `self._is_multi_fidelity`, right? Is there a reason to have the three pairs of variables?

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,605 @@
...
+        self._fidelity_confidences = copied_confidences
+
+        # Optional protocol for custom kernels that need access to the mapping.
+        if self.covar_module is not None and hasattr(
+            self.covar_module, "update_confidences"
+        ):
+            cast(Any, self.covar_module).update_confidences(copied_confidences)
+
+    def fit(self, observations: Iterable[Observation]) -> None:
+        """Fit the surrogate from scratch on all provided observations.
```
alexhernandezgarcia, line 125: I think I would mention in the docstring that this method updates `self._train_X`, `self._train_Y` and `self._is_multi_fidelity`.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,605 @@
...
+            return t.view(1, 1)
+        if t.ndim == 1:
+            return t.unsqueeze(-1)
+        return t
+
+    def _parse_observations(
+        self,
+        observations: Iterable[Observation],
+    ) -> tuple[torch.Tensor, torch.Tensor, bool]:
+        """Converts generic Observations into BoTorch-ready tensors.
```
alexhernandezgarcia, line 466: "BoTorch-ready" means that the fidelities are appended as the final column of the inputs, right? Anything else? Might be worth mentioning this in the description of the method.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point. Fixed: c0b633b1988aa49b825e8519524eded60ec8f51f.

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,605 @@
...
+            mean_out = mean.tolist()
+            std_out = std.tolist()
+
+        return {
+            "mean": mean_out,
+            "std": std_out,
+            "posterior": posterior,
+        }
+
+    def state_dict(self) -> Optional[dict[str, torch.Tensor]]:
```
alexhernandezgarcia: How about renaming this to `get_state_dict()`? Or was the intention to make it a property? If so, use decorator `@property`?

alexhernandezgarcia: Also, I think this method is isn't used at the moment. I guess this is used "in the future"?

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -0,0 +1,605 @@
...
+        -------
+        result : Optional[dict[str, torch.Tensor]]
+            Fitted model state dictionary, or ``None`` if the model has not yet
+            been built.
+        """
+        if self.model is None:
+            return None
+        return self.model.state_dict()
+
+    def load_state_dict(self, state_dict: dict[str, torch.Tensor]) -> None:
```
alexhernandezgarcia, line 330: I believe this method is currently not used. Just confirming that it will be needed in the future :)

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Agreed. Fixed: c8a21591d574e5fe5d90dd9db509bde0b898aa32.

src/activelearning/surrogate/botorch_surrogate.py
alexhernandezgarcia: Thanks!

src/activelearning/surrogate/botorch_surrogate.py
alexhernandezgarcia: Thanks!

alexhernandezgarcia (dismissed):
Wonderful work, thanks! Apologies that it took me a while to finish the review.

I made a bunch of comments, but there is nothing important and would be happy to have it merged as is, or I can re-approve after any further minor changes.

src/activelearning/utils/types.py
alexhernandezgarcia: Thank you for considering this!

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Yes, I believe so!

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: I agree! Fixed: 2ec8839734836a88c4b7be1388c3402c44f66126

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Good point. Yes, that is what it means.

Fixed: 62abc1193df6edf07c7bed8d559b06323126a7bd.

src/activelearning/surrogate/botorch_surrogate.py
alexhernandezgarcia: Thanks for considering this :)

src/activelearning/surrogate/botorch_surrogate.py
alexhernandezgarcia: Perfect, thanks!

alexhernandezgarcia (approved):
Re-approving after additional minor changes.

PR #4, opened 2026-02-26
Add Branin/Hartmann oracles and hypercube sampler

pedromsferraz-mila:
This PR adds BoTorch-based multi-fidelity benchmark oracles and a generative sampler for continuous-domain active learning. 

New components:
- AugmentedFunctionOracle: Base oracle for BoTorch augmented multi-fidelity test functions. Handles fidelity cost/confidence setup, cost querying, and evaluation. Fidelity confidence is derived proportionally from cost.

- BraninOracle: 2D Augmented Branin benchmark with 3 fidelity levels.

- Hartmann6DOracle: 6D Augmented Hartmann benchmark with 3 fidelity levels.

- HypercubeSampler: A generative sampler designed for continuous bounded domains where candidate pools are undefined. It draws samples from a bounded hypercube using either uniform or Latin Hypercube Sampling (LHS) strategies. For fidelity assignments, it supports both uniform sampling and strategies sampling inversely proportionally to cost.


src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,130 @@
...
+        function: SyntheticTestFunction,
+        fidelity_costs: dict[int, float],
+    ) -> None:
+        super().__init__()
+        self.fidelity_costs = fidelity_costs
+        max_cost = max(fidelity_costs.values())
+        self.fidelity_confidences: dict[int, float] = {
+            fidelity: cost / max_cost for fidelity, cost in fidelity_costs.items()
+        }
+        self._function = function
```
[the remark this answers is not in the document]
pedromsferraz-mila: Input validation will be done at the config parsing, when we implement it.

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,130 @@
...
+            normalized_fidelity = self.fidelity_confidences[fidelity]
+
+            x_tensor = torch.tensor(
+                [*candidate.x, normalized_fidelity], dtype=torch.float64
+            ).unsqueeze(0)
+            y_value = self._function(x_tensor).item()
+            observations.append(
+                Observation(x=candidate.x, y=y_value, fidelity=fidelity)
+            )
+        return observations
```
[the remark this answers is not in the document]
pedromsferraz-mila: Good point, fixed: [15cba1a](https://github.com/milaforscience/activelearning/pull/4/commits/15cba1acd531b6c99961771e45243ee512d05169)

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Uses the negated AugmentedHartmann from BoTorch, making all values positive.
+    Fidelity levels {1, 2, 3} are mapped to the [0, 1] fidelity range, with
+    level 3 corresponding to the full-fidelity Hartmann (s=1.0).
+    """
+
+    def __init__(self) -> None:
+        super().__init__(
+            # negate=True flips the naturally-negative Hartmann to positive values
+            function=AugmentedHartmann(negate=True),
```
olutayo-mila, line 159: It could be useful to expand the comment a bit to make it clearer why negate=True is set. Our active learning loop is formulated as a maximization problem, so negating the Hartmann function ensures consistency with the acquisition logic.

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,137 @@
...
+    ) -> None:
+        super().__init__()
+        if not fidelity_costs:
+            raise ValueError("fidelity_costs must not be empty")
+        self.fidelity_costs = fidelity_costs
+        max_cost = max(fidelity_costs.values())
+        self.fidelity_confidences: dict[int, float] = {
+            fidelity: cost / max_cost for fidelity, cost in fidelity_costs.items()
+        }
+        self._function = function
```
olutayo-mila: Minor suggestion: since we’re following Alex’s paper, the default linear confidence = cost / max_cost mapping makes sense here. That said, we could optionally accept a fidelity_confidences argument (defaulting to None) and fall back to the current behavior when it’s not provided. This would preserve the default while allowing flexibility in cases where confidence isn’t strictly proportional to cost.


```suggestion
    def __init__(
        self,
        function: SyntheticTestFunction,
        fidelity_costs: dict[int, float],
        fidelity_confidences: dict[int, float] | None = None,
    ) -> None:
        super().__init__()
        if not fidelity_costs:
            raise ValueError("fidelity_costs must not be empty")

        self.fidelity_costs = fidelity_costs

        if fidelity_confidences is None:
            max_cost = max(fidelity_costs.values())
            fidelity_confidences = {
                fidelity: cost / max_cost
                for fidelity, cost in fidelity_costs.items()
            }

        self.fidelity_confidences = fidelity_confidences
        self._function = function
```

pedromsferraz-mila: Good point, thanks for the suggestion!

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: Good point! Fixed: [4eed773](https://github.com/milaforscience/activelearning/pull/4/commits/4eed7731e3f7d4b562422ec80c280e9fcdcbad85)

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+    dimension is a fidelity parameter in [0, 1].
+
+    Subclasses only need to supply the underlying function instance and a
+    mapping of integer fidelity levels to costs.
+
+    Parameters
+    ----------
+    function : SyntheticTestFunction
+        An instantiated BoTorch test function whose last input dimension is
+        the fidelity parameter.
```
alexhernandezgarcia: Does the input function absolutely need to be a BoTorch `SyntheticTestFunction`? What exactly is the dependence on BoTorch?

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+    fidelity_costs : dict[int, float]
+        Mapping from integer fidelity level to query cost. The level with the
+        highest cost is treated as full fidelity (confidence = 1.0).
+    """
+
+    def __init__(
+        self,
+        function: SyntheticTestFunction,
+        fidelity_costs: dict[int, float],
+        fidelity_confidences: Optional[dict[int, float]] = None,
```
alexhernandezgarcia, line 37: This argument is not documented in the docstring. I think it'd be important to document in particular the behaviour when `fidelity_confidences` is None.

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+    ) -> None:
+        super().__init__()
+        if not fidelity_costs:
+            raise ValueError("fidelity_costs must not be empty")
+
+        self.fidelity_costs = fidelity_costs
```
alexhernandezgarcia: Should we check whether the fidelity indices are correct?

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+
+        if fidelity_confidences is None:
+            max_cost = max(fidelity_costs.values())
+            fidelity_confidences = {
+                fidelity: cost / max_cost for fidelity, cost in fidelity_costs.items()
+            }
+
+        self.fidelity_confidences = fidelity_confidences
```
alexhernandezgarcia: Should we validate the input confidences and normalise them?

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+        self._function = function
+
+    def _validate_fidelity(self, candidate: Candidate) -> int:
```
alexhernandezgarcia: Could this method not be part of the parent class `Oracle`?

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+
+import torch
+from botorch.test_functions.multi_fidelity import AugmentedBranin, AugmentedHartmann
+from botorch.test_functions.synthetic import SyntheticTestFunction
+
+from activelearning.oracle.oracle import Oracle
+from activelearning.utils.types import Candidate, Observation
+
+
+class AugmentedFunctionOracle(Oracle):
```
alexhernandezgarcia: Why not inheriting from `MultiFidelityOracle`?

alexhernandezgarcia: I have the impression that this class reimplements part of the functionality in `MultiFidelityOracle` in slightly different ways.

src/activelearning/oracle/augmented_function_oracle.py
alexhernandezgarcia: (Some of this functionality is already in `MultiFidelityOracle`)

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Returns
+        -------
+        observations : list[Observation]
+            Observations with the function value as y.
+        """
+        if not candidates:
+            return []
+
+        fidelities = [self._validate_fidelity(c) for c in candidates]
```
alexhernandezgarcia: What do you think about packing this into a method `get_fidelities(self, candidates: Sequence[Candidate])`?

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+
+        rows = [
+            [*c.x, self.fidelity_confidences[f]] for c, f in zip(candidates, fidelities)
+        ]
+        x_batch = torch.tensor(rows, dtype=torch.float64)
```
alexhernandezgarcia: Are we always going to use `torch.float64` everywhere? Or should this be a config input? (Happy to have this discussion outside this PR)

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -0,0 +1,177 @@
...
+        If None, confidences are derived proportionally from costs.
+    """
+
+    def __init__(
+        self,
+        fidelity_costs: dict[int, float],
+        fidelity_confidences: Optional[dict[int, float]] = None,
+    ) -> None:
+        super().__init__(
+            function=AugmentedHartmann(negate=True),
```
alexhernandezgarcia, line 159: Could add here a similar comment as for Branin.

src/activelearning/sampler/hypercube_uniform_sampler.py
```diff
@@ -0,0 +1,103 @@
...
+    Raises
+    ------
+    ValueError
+        If any lower bound >= upper bound or num_samples <= 0.
+    """
+
+    def __init__(
+        self,
+        bounds: Sequence[tuple[float, float]],
+        num_samples: int,
```
alexhernandezgarcia: I think I had this thought while reviewing PR https://github.com/milaforscience/activelearning/pull/2 but I didn't make this comment: what was the reason to not have `num_samples` as an input to `sample()` and have it instead as a class argument?

src/activelearning/sampler/hypercube_uniform_sampler.py
```diff
@@ -0,0 +1,103 @@
...
+        fidelities: Optional[Sequence[int]] = None,
+    ) -> None:
+        if num_samples <= 0:
+            raise ValueError(f"num_samples must be > 0, got {num_samples}")
+        for i, (lower, upper) in enumerate(bounds):
+            if lower >= upper:
+                raise ValueError(f"Bound {i} has lower >= upper: ({lower}, {upper})")
```
alexhernandezgarcia: Changing `i` to `idx` 😅 (feel free to ignore though)
```suggestion
        for idx, (lower, upper) in enumerate(bounds):
            if lower >= upper:
                raise ValueError(f"Bound {idx} has lower >= upper: ({lower}, {upper})")
```

src/activelearning/sampler/hypercube_uniform_sampler.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if fidelities is not None and len(fidelities) == 0:
+            raise ValueError("fidelities must not be empty when specified")
+        self.bounds = bounds
+        self.num_samples = num_samples
+        self.fidelities = list(fidelities) if fidelities is not None else None
+
+        lowers = [b[0] for b in bounds]
+        uppers = [b[1] for b in bounds]
```
alexhernandezgarcia: One-liner alternative - but I'm with the current way too
```suggestion
        lowers, uppers = zip(*[b for b in bounds])
```

alexhernandezgarcia: Or even this, to avoid looping again over `lowers` and `uppers` in line 62:

```
lowers, uppers, diff = zip(*[(*b, b[1] - b[0]) for b in bounds])
```

src/activelearning/sampler/hypercube_uniform_sampler.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def sample(
+        self,
+        acquisition: Optional[Acquisition] = None,
+        observations: Optional[Iterable[Observation]] = None,
+    ) -> list[Candidate]:
+        """Generate candidates by sampling uniformly from the hypercube.
+
+        Each candidate's fidelity is drawn uniformly at random from the
+        ``fidelities`` list supplied at construction.
```
alexhernandezgarcia: Sampling (inversely) proportionally to the cost may be a more natural baseline, but we can implement this later.

src/activelearning/sampler/hypercube_uniform_sampler.py
```diff
@@ -0,0 +1,103 @@
...
+            Unused. Present for interface compatibility.
+
+        Returns
+        -------
+        result : list[Candidate]
+            ``num_samples`` candidates with ``x`` as a plain Python list of
+            floats and ``fidelity`` drawn uniformly from ``fidelities``.
+        """
+        # Shape: (num_samples, n_dims)
+        uniform = torch.rand(self.num_samples, len(self.bounds), dtype=torch.float64)
```
alexhernandezgarcia: If we are converting to list later, would not make more sense to do all the random sampling with `random` or numpy and bypass torch?

alexhernandezgarcia (commented):
Nice progress!

My main question is about the decision of not using `MultiFidelityOracle` and its interface with `fidelity_configs`. I think the interface implemented here might be clearer and more general, but in this case, I would consider changing `MultiFidelityOracle` and try to avoid two parallel implementations for a very similar thing. It's also possible I may be missing something 😅 

The other question or suggestion is about avoiding using torch if the tensors will immediately be converted to Python lists.

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: BoTorch provides [multi-fidelity augmented test functions](https://botorch.readthedocs.io/en/latest/test_functions.html#module-botorch.test_functions.multi_fidelity) (`AugmentedBranin`, `AugmentedHartmann`), and since BoTorch is already a dependency for the GP surrogate, it made sense to reuse them rather than reimplement the math. `AugmentedFunctionOracle` is intentionally a thin wrapper around these functions, with `BraninOracle` and `Hartmann6DOracle` as its only concrete subclasses in this PR.

The dependence on `SyntheticTestFunction` is not a hard technical requirement. At runtime, only `__call__` is used, so any callable accepting a batched input tensor with fidelities concatenated as last dimension would work. The oracle itself is responsible for constructing that tensor (appending the fidelity as the last dimension). However, using `SyntheticTestFunction` as the type annotation tightens the contract to BoTorch's test function hierarchy, documents intent, and leaves the door open to using dim, bounds, or other metadata in the future.

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: Good catch, thanks! 93997d2c9617dd7b149742bc5ddac9149289e1d3

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: Good point, I've moved the validation code to the base `Oracle` class and am using it in both classes now: c966a68e1c4d2fb3ab2954ab02378ae581b6a00a

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: Good point, I've moved the validation methods to the parent `Oracle` class: c966a68e1c4d2fb3ab2954ab02378ae581b6a00a

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: Great point, there is definitely some shared logic here. I initially considered inheritance but opted against it because the interfaces are too divergent:

- `MultiFidelityOracle` relies on discrete fidelity levels, each mapped to a specific `score_fn`.
- `AugmentedFunctionOracle` supports continuous fidelity in $[0, 1]$ via a single function call. It expects a batched tensor where fidelity is the final dimension, maintaining compatibility with BoTorch synthetic functions (e.g., AugmentedBranin and AugmentedHartmann).

Inheriting would require overriding nearly every core method just to access the validation logic, which creates fragile, "forced" coupling. Instead, I’ve promoted the shared fidelity and confidence validation to the `Oracle` base class. This keeps the implementation DRY without the inheritance overhead. Let me know what you think!

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: Good point, added some validation: 3f70c89d95f06fb859bb00b86434c492b01124fa

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: I think I'd prefer to keep this inline. it validates each candidate and collects fidelities; extracting it would only relocate coupled logic without improving readability. Let me know what you think.

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: I’ve been thinking about this as well. I believe the best would be to have this at a top-level config field. Since we don't have config management yet, I've hardcoded it as `torch.float64` for now, but I'll make sure to migrate this over in the next PR once the config parsing is in place.

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: Good point, added: 9fab78ae8e5e8ce20d6b7b97cb738fb8f3853fda

src/activelearning/sampler/hypercube_uniform_sampler.py
pedromsferraz-mila: I chose to implement `num_samples` as a constructor-level setting rather than an input to `sample()` to maintain a clean decoupling between the sampler and the active learning (AL) loop. 

If we made it a `sample()` argument, the AL loop would be forced to carry the `num_samples` parameter even though it's used only by the sampler, which leaks implementation details into the main loop. By keeping it in the constructor, the AL loop stays simplified and doesn't need to manage variables that don't interact with other objects in the process. 

Additionally, this approach provides better flexibility for the future; some samplers might implement strategies based on costs or other criteria rather than a fixed `num_samples`, and this structure allows for that sampler-specific behavior without forcing a rigid signature on the interface.

src/activelearning/sampler/hypercube_uniform_sampler.py
pedromsferraz-mila: Yes, good point!

src/activelearning/sampler/hypercube_uniform_sampler.py
pedromsferraz-mila: Good point! Fixed: c996060a30a9ab951d0641afd11fcee723ea4bd8

src/activelearning/sampler/hypercube_uniform_sampler.py
pedromsferraz-mila: With the native `random` library we wouldn't be able to use broadcasting, which would make the sampling logic more verbose. As for numpy, we would still have to convert the result from a `np.array` to a list at the end, so the conversion effort would be essentially the same.

With that in mind, I’d argue for keeping torch for the convenience and consistency across the repo, but let me know your thoughts!

src/activelearning/sampler/hypercube_uniform_sampler.py
pedromsferraz-mila: Good point! I've renamed it to `HypercubeSampler` and added the option to sample fidelities proportionally to the cost. I've also added the option to sample coordinates via Latin Hypercube Sampling (LHS) instead of uniformly, as was suggested by Gaétan. 

We now have different sampling strategies for both point generation and fidelity assignment, making it much more flexible as a baseline sampler.

src/activelearning/oracle/augmented_function_oracle.py
alexhernandezgarcia: Thanks for your answer!

I am bit confused: `MultiFidelityOracle` relies on discrete fidelity levels, but in practice Branin and Hartmann will also be used with discrete fidelity levels, right?

If I understand correctly, the issue here is that to use the fidelity feature of BoTorch's synthetic functions, we (quite annoyingly) have to append the fidelity confidence to the actual queries and that would break the logic of `MultiFidelityOracle`, for example in the method `query()`. Correct?

Would you agree that if that was not an issue, we then would be able to implement these oracles as `MultiFidelityOracle`'s? If so, it may be worth thinking of an alternative implementation that allows to stick to our planned class for multi-fidelity oracles, no?

One solution could be to entirely bypass BoTorch for this. It's not that the [Branin](https://botorch.readthedocs.io/en/stable/_modules/botorch/test_functions/multi_fidelity.html#AugmentedBranin) and [Hartmann](https://botorch.readthedocs.io/en/stable/_modules/botorch/test_functions/multi_fidelity.html#AugmentedHartmann) functions are some kind of obscure code that we can only access through BoTorch 😅 

Another solution would be to implement some kind of wrapper to still use BoTorch augmented functions while using `MultiFidelityOracle`. I haven't exhaustively thought about what that would entail, but I'm imagining it would not be too much.

Do you think it's worth thinking a bit more about this, or would you just go ahead with this implementation. Also sorry if I am missing something important.

src/activelearning/oracle/augmented_function_oracle.py
alexhernandezgarcia: That makes sense, thanks for explaining. I would refer to [my other comment](https://github.com/milaforscience/activelearning/pull/4/changes#r2913366323) about the possibility of inheriting `MultiFidelityOracle` but this comment can be resolved.

src/activelearning/oracle/augmented_function_oracle.py
alexhernandezgarcia: This seems good to me 👍

src/activelearning/oracle/augmented_function_oracle.py
alexhernandezgarcia: Sounds good, I'm fine leaving it hard-coded here. I just wanted to get that conversation started and I agree :)

src/activelearning/oracle/augmented_function_oracle.py
alexhernandezgarcia: Ok, I agree, thanks!

src/activelearning/sampler/hypercube_uniform_sampler.py
alexhernandezgarcia: Perfect, thanks for explaining. I agree :+1:

src/activelearning/sampler/hypercube_uniform_sampler.py
alexhernandezgarcia: It wasn't "broken" but thanks! ;P See [this comment](https://github.com/milaforscience/activelearning/commit/c996060a30a9ab951d0641afd11fcee723ea4bd8#r179200315).

src/activelearning/sampler/hypercube_uniform_sampler.py
alexhernandezgarcia: Fantastic, thanks! I didn't know the name Latin Hypercube Sampling but that's great to have and I have learnt something :)

src/activelearning/sampler/hypercube_uniform_sampler.py
alexhernandezgarcia: Ok thanks, fine with me!

alexhernandezgarcia (dismissed):
Thanks for addressing all my comments! I have no further remaining comments beyond the conversation about a potential re-implementation of Branin and Hartmann using `MultiFidelityOracle` instead a new class. I'd be happy to have this merged, since it's been reviewed, and continue that conversation outside this PR.

src/activelearning/sampler/hypercube_uniform_sampler.py
pedromsferraz-mila: Thanks for the comment, I've removed the additional unused field!

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: Good points! After our discussion and reading your comments, I agree it is cleaner to implement this as just a wrapper on top of `MultiFidelityOracle`. I've made the changes on [this commit](https://github.com/milaforscience/activelearning/pull/4/commits/454eea1e583d9992e7bad38a23f2e0b09bb66e8a), let me know what you think!

alexhernandezgarcia: I agree this is cleaner, thanks for the changes!

alexhernandezgarcia (approved):
Re-approved!

PR #5, opened 2026-03-17
Add loggers and runtime context

pedromsferraz-mila:
This PR introduces a shared logging interface with multiple backends: `ConsoleLogger`, `WandbLogger`, `CometLogger`, and `AimLogger`, plus `MultiLogger` for fan-out to multiple destinations.

It also adds `RuntimeContext` and `ALRuntimeMixin`, giving active learning components a shared way to access runtime services such as `logger`, `device`, `dtype`, and `precision`. This makes it easy for classes to log metrics and figures without coupling their logic to a specific backend.

To demonstrate the new logging flow and provide useful diagnostics, this PR also adds Branin landscape plotting/logging, so runs can emit visualizations of queried points over the objective landscape.


src/activelearning/sampler/hypercube_sampler.py
```diff
@@ -101,19 +109,17 @@ def _generate_points(self) -> torch.Tensor:
...
+    def _latin_hypercube(self, num_samples: int, n_dims: int) -> torch.Tensor:
+        """Generate a Latin Hypercube Sample in the unit hypercube ``[0, 1]^d``.
 
         LHS ensures that each dimension is divided into 'num_samples' equal-width bins,
         and each bin contains exactly one sample. This provides better space-filling
         properties than simple random sampling.
         """
         # 1. Generate jitter/offsets: rand(0, 1) for each sample's position within its bin.
-        offsets = torch.rand(num_samples, n_dims, dtype=torch.float64)
+        offsets = torch.rand(num_samples, n_dims, dtype=self.dtype)
```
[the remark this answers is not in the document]
pedromsferraz-mila: In the current implementation, the tensors will be converted to a Python list at the end of the `sample` call, so there's no need to move them to CUDA devices.

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -119,13 +132,39 @@ def __init__(
...
+            fidelity_confidences=self.get_fidelity_confidences(),
+            supported_fidelities=self.get_supported_fidelities(),
+            dtype=self.dtype,
+            device=self.device,
+            title="Branin landscape with queried candidates",
+        )
+        try:
+            self.logger.log_figure("branin_landscape_query", figure)
+        finally:
+            plt.close(figure)
```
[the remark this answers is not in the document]
pedromsferraz-mila: Good point, I've added a flag to toggle landscape logging on/off: f2d40d5761d4426de1276f26ec632e9eb4a40e32

src/activelearning/logger/logger.py
```diff
@@ -0,0 +1,499 @@
...
+    run_name : str, optional
+        Name of this specific run. Defaults to a timestamp if not provided.
+    """
+
+    def __init__(
+        self, project_name: str, run_name: str | None = None, **kwargs: Any
+    ) -> None:
+        self.project_name = project_name
+        if run_name is None:
+            run_name = datetime.today().strftime("%d/%m-%H:%M:%S")
```
alexhernandezgarcia: In the past, with the gflownet repo, we've gone through headaches a couple of times due to multiple runs sharing the same name, when they are scheduled at the same time in a cluster. To spare ourselves these headaches in the future, a solution is adding microseconds to the name with `%f`. See here, for example: https://github.com/alexhernandezgarcia/gflownet/blob/main/config/train.yaml#L27

src/activelearning/logger/logger.py
```diff
@@ -0,0 +1,499 @@
...
+    """
+
+    def __init__(
+        self, project_name: str, run_name: str | None = None, **kwargs: Any
+    ) -> None:
+        super().__init__(project_name, run_name, **kwargs)
+        import wandb  # type: ignore[import-not-found]  # optional dependency
+
+        self._wandb = wandb
+        self.run = wandb.init(project=project_name, name=self.run_name)
```
alexhernandezgarcia: I think it would change anything now, but I think it seems that this should be `project=self.project_name` for consistency, right?

src/activelearning/logger/logger.py
pedromsferraz-mila: Good point, I've had similar issues in the past haha

Fixed! [33f9fc7](https://github.com/milaforscience/activelearning/pull/5/commits/33f9fc7a0b49342b6c49ee52b6ff41885a77b778)

src/activelearning/logger/logger.py
pedromsferraz-mila: Good point, fixed for other loggers as well! [19ff9dd](https://github.com/milaforscience/activelearning/pull/5/commits/19ff9dd9936b8cd23e2bb4a949f53c2a4e9625be)

src/activelearning/logger/logger.py
alexhernandezgarcia: I'm not suprised :sweat_smile: Thanks for the modification!

src/activelearning/logger/logger.py
```diff
@@ -0,0 +1,500 @@
...
+        run_name: str | None = None,
+        workspace: str | None = None,
+        api_key: str | None = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(project_name, run_name, **kwargs)
+        import comet_ml  # type: ignore[import-not-found]  # optional dependency
+
+        self._comet_ml = comet_ml
+        self.experiment = comet_ml.Experiment(
```
alexhernandezgarcia, line 264: In WandbLogger the run/experiment is stored as `self.run` and here in CometLogger it is `self.experiment`. Is it intentional, perhaps to stick to each framework terminology?

src/activelearning/logger/logger.py
```diff
@@ -0,0 +1,500 @@
...
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(project_name, run_name, **kwargs)
+        import aim  # type: ignore[import-not-found]  # optional dependency
+
+        self._aim = aim
+        self.run = aim.Run(repo=repo, experiment=self.project_name)
+        self.run.name = self.run_name
+        self._buffer: dict[str, Any] = {}
+        self._current_step: int = 0
```
alexhernandezgarcia: I have the impression that `self._current_step` is not truly used. It is updated in `log_step()` but not really used.

src/activelearning/runtime.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@dataclass
+class RuntimeContext:
+    """Shared runtime settings available to active learning components."""
+
+    logger: Logger | None = None
+    device: torch.device = torch.device("cpu")
+    dtype: torch.dtype = torch.float64
+    precision: Literal[32, 64] = 64
```
alexhernandezgarcia: Why do we need both dtype and precision? Would things work correctly if one uses for example dtype = torch.float64 and precision = 32?

src/activelearning/oracle/augmented_function_oracle.py
```diff
@@ -111,21 +124,53 @@ class BraninOracle(AugmentedFunctionOracle):
...
+            evaluator=self._function,
+            candidates=candidates,
+            bounds=((-5.0, 10.0), (0.0, 15.0)),
+            fidelity_confidences=self.get_fidelity_confidences(),
+            supported_fidelities=self.get_supported_fidelities(),
+            dtype=self.dtype,
+            device=self.device,
+            title="Branin landscape with queried candidates",
+        )
+        try:
```
alexhernandezgarcia, line 169: Why is `try` needed?

alexhernandezgarcia (dismissed):
Very neat! It's amazing to have all these loggers and a neat solution for passing the context around.

I haven't reviewed the tests and the plotting methods in much detail.

I dropped a few comments but they're all minor so I'm approving the PR. Happy to re-approve as quickly as I can if there are additional minor changes.

One suggestion: at this point, it may be already a good idea to add a few words about how to run the code in the README.

src/activelearning/logger/logger.py
pedromsferraz-mila: That's right, Wandb uses a [Run](https://docs.wandb.ai/models/ref/python/experiments/run) object, while Comet uses an [Experiment](https://www.comet.com/docs/v2/api-and-sdk/python-sdk/reference/Experiment/) object, so I've decided to use names consistent with their own framework. 

This also makes the code easier to read for people that already have experience with that framework (e.g., in Comet, it's standard to write `self.experiment.log_<...>`).

src/activelearning/logger/logger.py
pedromsferraz-mila: Good catch, thanks! Fixed: fa2bb97283e316b66acd17290205f41cb574d08d

src/activelearning/runtime.py
pedromsferraz-mila: Good catch! I originally added precision to keep the API backend-agnostic (e.g. torch vs numpy), but that also introduced redundancy with dtype and allowed inconsistent combinations. Also, dtype is the more expressive setting, since it can distinguish cases like torch.float16 vs torch.bfloat16, which a plain precision value cannot. 

I’ve updated the code to remove precision and use dtype as the single source of truth, let me know if that works for you! 2ab8f1812ce8f36412bd381965d11853a0f2f2c9

src/activelearning/oracle/augmented_function_oracle.py
pedromsferraz-mila: It's to prevent Matplotlib memory leaks. If `self.logger.log_figure()` throws an exception, the `finally` block guarantees that `plt.close(figure)` still runs, ensuring the figure is properly cleared from memory.

src/activelearning/logger/logger.py
alexhernandezgarcia: Thanks for explaining, that makes sense!

src/activelearning/runtime.py
alexhernandezgarcia: Ok I see, yes the update seems good to me :+1:

src/activelearning/oracle/augmented_function_oracle.py
alexhernandezgarcia: Oh good catch! I had not thought about that!

alexhernandezgarcia (approved):
Re-approving after minor changes addressing the comments.

PR #6, opened 2026-03-20
Add exact knapsack-based selector

pedromsferraz-mila:
This PR introduces `KnapsackSelector`, which formulates candidate selection as a 0/1 knapsack problem and uses PuLP/CBC to maximize acquisition value within a fixed budget.

It also adds `ClusteredKnapsackSelector`, a subclass that extends the MIP with per-cluster cardinality constraints. It clusters the candidate pool (via `scikit-learn`’s K-means or DBSCAN) and limits selections per cluster to a configurable max_per_cluster. This encourages batch diversity without sacrificing the optimality of the knapsack solution.

Additionally, this PR includes an optional greedy warm start, a helper for value-to-cost ranking, and comprehensive unit tests.


src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # If acquisition scores are all zero, set constant score to pick cheapest items
+        if all(v == 0 for v in acq_values):
+            acq_values = [1.0] * len(candidates)
+
+        # Initialize the Maximization problem
+        prob = pulp.LpProblem("Knapsack", pulp.LpMaximize)
+
+        # Define Decision Variables (0 or 1 for each item)
+        n = len(costs)
```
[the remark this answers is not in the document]
pedromsferraz-mila: Unnecessary, as this is currently enforced by the interface for acquisition and cost_fn.

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,203 @@
...
+            raise ValueError("Cost function is required for KnapsackSelector.")
+        if round_budget is None:
+            raise ValueError("Budget is required for KnapsackSelector.")
+        if not candidates:
+            return []
+
+        # Acquisition scores should be non-negative
+        acq_values = [max(0, v) for v in acquisition.score(candidates)]
+        costs = cost_fn(candidates)
+
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [fcaf2d0](https://github.com/milaforscience/activelearning/pull/6/commits/fcaf2d01d07b8a61ab69d5bb22da1366b6c6b992)

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,203 @@
...
+        max_value = pulp.value(prob.objective)
+        if max_value is None:
+            raise ValueError("No feasible solution found within time limit.")
+
+        if pulp.LpStatus[status] != "Optimal":
+            print(
+                f"Warning: Optimization ended with status '{pulp.LpStatus[status]}'. "
+                "The selected solution may not be optimal."
+            )
+
```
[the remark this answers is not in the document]
pedromsferraz-mila: Will change this once logging is implemented.

tests/selector/test_knapsack_selector.py
```diff
@@ -0,0 +1,105 @@
...
+    assert selected_indices == [0, 2]
+
+
+@pytest.mark.parametrize("warm_start", [False, True])
+def test_knapsack_selector_finds_exact_solution_when_greedy_misses(warm_start):
+    """Test exact knapsack solve beats the greedy warm-start heuristic."""
+    candidates = [Candidate(x=0), Candidate(x=1)]
+    acquisition = Mock()
+    acquisition.score.return_value = [10.0, 9.0]
+
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [fcaf2d0](https://github.com/milaforscience/activelearning/pull/6/commits/fcaf2d01d07b8a61ab69d5bb22da1366b6c6b992)

tests/selector/test_knapsack_selector.py
```diff
@@ -0,0 +1,105 @@
...
+    monkeypatch.setattr(
+        knapsack_selector_module,
+        "greedy_knapsack_indices",
+        recording_helper,
+    )
+
+    acquisition = Mock()
+    acquisition.score.return_value = [8.0, 5.0]
+
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [fcaf2d0](https://github.com/milaforscience/activelearning/pull/6/commits/fcaf2d01d07b8a61ab69d5bb22da1366b6c6b992)

tests/selector/test_knapsack_selector.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+
+    assert len(helper_calls) == expected_calls
+
+
+def test_knapsack_selector_returns_empty_for_empty_candidates():
+    """Test selector returns an empty list when the candidate pool is empty."""
+    acquisition = Mock()
+    acquisition.score.return_value = []
+    selector = KnapsackSelector()
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [fcaf2d0](https://github.com/milaforscience/activelearning/pull/6/commits/fcaf2d01d07b8a61ab69d5bb22da1366b6c6b992)

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,203 @@
...
+        selected_candidates = [
+            candidates[i] for i in range(n) if pulp.value(x[i]) > 0.5
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: 7a7c9c5ddbd776818f18393334f8eee8c76da2d7

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Solve using the default CBC solver (included with PuLP)
+        status = prob.solve(
+            pulp.PULP_CBC_CMD(
+                timeLimit=self.time_limit,
+                msg=self.verbose,
+                warmStart=self.warm_start,
+            )
+        )
+
```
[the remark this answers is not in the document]
pedromsferraz-mila: Good point. Even though CBC usually comes bundled with Pulp, it's good to check for robustness. [86d411a](https://github.com/milaforscience/activelearning/pull/6/commits/86d411ada53f9688191adebf064692e19efde35c)

src/activelearning/selector/clustered_knapsack_selector.py
```diff
@@ -0,0 +1,181 @@
...
+                for c in candidates
+            ]
+        )
+
+        if self.clustering == "kmeans":
+            from sklearn.cluster import KMeans  # noqa: PLC0415
+
+            labels: np.ndarray = KMeans(
+                n_clusters=self.n_clusters, n_init="auto"
+            ).fit_predict(features)
```
olutayo-mila, line 175: Right now KMeans(n_clusters=self.n_clusters).fit_predict(features) is called without checking len(candidates) >= self.n_clusters. If fewer candidates are passed than clusters, sklearn raises a raw ValueError: n_samples=X should be >= n_clusters=Y but perhaps it might be better to add a clear library error.


```suggestion
if self.clustering == "kmeans" and len(candidates) < self.n_clusters:
     raise ValueError(
         f"Cannot cluster {len(candidates)} candidates into {self.n_clusters} "
         f"clusters. Provide at least {self.n_clusters} candidates or reduce n_clusters."
     )
```

olutayo-mila (dismissed):
Nice work! Just had one really minor comment.

src/activelearning/selector/clustered_knapsack_selector.py
pedromsferraz-mila: Good point, that'd be easier to understand! Fixed: https://github.com/milaforscience/activelearning/pull/6/commits/cf0e91906a1e32d8f5df360ac3be2595c6d73895

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,262 @@
...
+from activelearning.utils.types import Candidate
+
+
+class KnapsackSelector(Selector):
+    """Selector that solves a 0/1 knapsack problem over candidate utilities.
+
+    Parameters
+    ----------
+    time_limit : float, optional
+        Optional CBC time limit in seconds.
```
carriepl: What is CBC?

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,262 @@
...
+    verbose : bool, default=False
+        Whether to emit solver logs.
+    warm_start : bool, default=False
+        Whether to seed CBC with the greedy knapsack solution before solving the
+        exact mixed-integer program.
```
carriepl: From a user perspective, how do I know whether or not I should use warm_start? From the description, it sounds like it should help but from the default value (False) is sounds like it's mostly not a good thing or, at the very least, it has very unreliable effects.

If you have some wisdom to impact to the future user regarding how to use this, I think that it is worth putting it somewhere in the docstring.

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+        if cost_fn is None:
+            raise ValueError(f"Cost function is required for {type(self).__name__}.")
+        if round_budget is None:
+            raise ValueError(f"Budget is required for {type(self).__name__}.")
+        if not candidates:
+            return []
+
+        # Acquisition scores should be non-negative
+        acq_values = [max(0, v) for v in acquisition.score(candidates)]
```
carriepl: (sorry about that, this comment exploded. I wrote the first part and continued on but something was gnawing at me so I came back and wrote a second part)

Part 1 : the original comment
---------------------------------------

If the Acquisition object gives negative scores, this will squish them all together and potentially lose quite a bit of information (depending on how the Acquisition object works).

I don't recall seeing a constraint that scores have to be >=0 anywhere else in the framework so someone implementing their own Acquisition object might do something that the framework doesn't explicitely discourage (return negative scores) and it might work for most selectors and then it might completely break for this one.

Part 2 : the sequel
---------------------------------------

And doing *that* requires opening a can of worms and clarifying the semantics of the acquisition scores and the semantics of your knapsack selector class. What do the acquisition scores mean, exactly? And what do the score-aware selectors do exactly?

So far, I my understanding of the framework, my interpretation of the acquisition scores was just that "higher = better, lower = worse". There's a monotonic relationship between scores and quality but the just a single score by itself doesn't tell you anything. A negative acquisition score isn't necessarily "bad" : it doesn't mean that a candidate has "negative value", it just means that this candidate has less value than another candidate that would have an acquisition score of 0. That works as long as the task of your selectors is to keep the best candidates withing the available budget.

The problem with that is that the knapsack problem makes a different set of incompatible assumptions.  It doesn't pick the best candidates, it tries to maximise the total value. So in a knapsack problem, you don't want items with negative values. Even if you still have room, you don't take them.

Sooooo.... I guess one of two things : 
1- The framework needs to establish, and enforce, stronger conventions to govern acquisition scores. Instead of "higher = better", there needs to be an explicit framework position on negative acquisition scores. Either "negative = bad" (negative values mean that labeling a candidate is estimated to be bad), or "negative = forbidden" (Acquisition objects can't return negative values)
OR
2- If the acquisition score keep the same semantics as up until now, the KnapsackSelector needs to make very *very* explicit that it make an important assumption that the rest of the framework doesn't make. Docstrings, comments, etc. I think that it probably should even raise errors if an acquisition score is negative since that likely means that the user didn't know and/or didn't pay attention (although you might add an argument to ignore this error, with a default value of False to avoid overeagerly hiding errors from users). 

Though it requires more work, I think that no1 is the better long-term solution of the two. I think that even with good documentation, no2 can remain confusing to users and to people who develop either Acquisition or Selector objects.

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,262 @@
...
+
+        # Solve using the default CBC solver (included with PuLP)
+        solver = self._build_solver()
+        status = prob.solve(solver)
+        variable_values = self._extract_solution_values(x, status)
+
+        selected_candidates = [
+            candidate
+            for candidate, variable_value in zip(candidates, variable_values)
+            if variable_value > 0.5
```
carriepl, line 157: Maybe a comment to explain here. Does pulp not return binary variables?

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,262 @@
...
+        ]
+        return selected_candidates
+
+    def _add_extra_constraints(
+        self,
+        prob: pulp.LpProblem,
+        x: list[pulp.LpVariable],
+        candidates: Sequence[Candidate],
+    ) -> None:
+        """Hook for subclasses to inject extra constraints into the MIP.
```
carriepl: Not everyone who is going to read this is going to know what a mixed-integer program is. Maybe add an explanation, or simply avoir mentioning the MIP (since it's not required to understand the code). For instance  "[...] to inject extra constraints into the knapsack optimization problem" or anything else.

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,262 @@
...
+        runs.  The default implementation is a no-op.
+
+        Parameters
+        ----------
+        prob : pulp.LpProblem
+            The in-progress maximisation problem.
+        x : list[pulp.LpVariable]
+            Binary decision variables aligned with *candidates*.
+        candidates : Sequence[Candidate]
+            The same candidate pool passed to ``__call__``.
```
carriepl, line 179: Would be worth having a link to documentation, or an example, or something like that.

src/activelearning/selector/knapsack_selector.py
```diff
@@ -0,0 +1,262 @@
...
+        # Constraint: Total cost must be <= budget
+        prob += pulp.lpSum([costs[i] * x[i] for i in range(n)]) <= round_budget
+
+        # Allow subclasses to add extra constraints
+        self._add_extra_constraints(prob, x, candidates)
+
+        # Solve using the default CBC solver (included with PuLP)
+        solver = self._build_solver()
+        status = prob.solve(solver)
+        variable_values = self._extract_solution_values(x, status)
```
carriepl: Since it's possible to set a time limit, is there any chance that the time limit will be reached before even a first solution is found? Is there a guarantee that, whenever you stop, there will be a valid (though perhaps not optimal) solution to extract?

tests/selector/test_knapsack_selector.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def test_greedy_knapsack_indices_breaks_ratio_ties_by_value_then_index():
+    """Test greedy helper preserves the documented tie-break order."""
+    values = [10.0, 8.0, 10.0]
+    costs = [5.0, 4.0, 5.0]
+
+    selected_indices = greedy_knapsack_indices(values, costs, budget=10.0)
+
+    assert selected_indices == [0, 2]
```
carriepl, line 82: Maybe add a test to ensure that it does behave greedily. Essentially a case where is will miss the optimal solution in favor of the greedy one.

src/activelearning/selector/knapsack_selector.py
carriepl: Ok, I just got later in the code and saw that this point is handled. However, I have a follow up question but it's a bit more design-oriented than code-oriented.

How should the framework handle the case where no sample can be selected (say, because the budget is insufficient). Select none? Throw exception and then crash and burn?

On one hand, it might be indicative that something is wrong with the experiment parameters (hence a valid reason to throw an exception). However, it is also possible that it happens because of a fluke. When you have costs that depend on the candidate, maybe through some fluke of probability, your GFN sampler only gives you super expensive samples of the highest fidelity and you can't afford any (but there's no way for the sampler to know because only the oracle can say what the real labeling cost will be). And if there is any chance of this happening naturally, the longer the experiments run, the more likely that it might happen purely by chance.

tests/selector/test_knapsack_selector.py
```diff
@@ -0,0 +1,223 @@
...
+
+    monkeypatch.setattr(knapsack_selector_module.pulp.LpProblem, "solve", fake_solve)
+
+    with pytest.raises(ValueError, match="without a usable incumbent solution"):
+        selector(
+            [Candidate(x=0), Candidate(x=1)],
+            acquisition=acquisition,
+            cost_fn=lambda _candidates: [1.0, 1.0],
+            round_budget=1.0,
+        )
```
carriepl, line 263: Would this be a valid reason to make a recommendation about the use of warm_start? To ensure that there is at least a solution at the end? Is there any reason not to use the warm start?

tests/selector/test_knapsack_selector.py
```diff
@@ -0,0 +1,223 @@
...
+
+    selected = selector(
+        [Candidate(x=0), Candidate(x=1)],
+        acquisition=acquisition,
+        cost_fn=lambda _candidates: [1.0, 1.0],
+        round_budget=1.0,
+    )
+
+    assert [candidate.x for candidate in selected] == [0]
+    assert "Using the best available solution found so far" in capsys.readouterr().out
```
carriepl, line 287: Can you also have a test for this use case where the early stop is triggered through an explicit use of the time limit on the solver?

tests/selector/test_clustered_knapsack_selector.py
```diff
@@ -0,0 +1,425 @@
...
+    selector = ClusteredKnapsackSelector(
+        max_per_cluster=1, clustering="kmeans", n_clusters=2
+    )
+    selected = selector(
+        candidates,
+        acquisition=_mock_acquisition([1.0, 1.0, 1.0, 1.0]),
+        cost_fn=_cost_fn([1.0, 1.0, 1.0, 1.0]),
+        round_budget=100.0,
+    )
+    assert len(selected) <= 2
```
carriepl: How would len(selected) == 1 be a valid test result, here? From the parameters defined, it seems like the solution will necessarily have to select 2 candidates.

tests/selector/test_clustered_knapsack_selector.py
```diff
@@ -0,0 +1,425 @@
...
+    selector = ClusteredKnapsackSelector(
+        max_per_cluster=1, clustering="dbscan", eps=0.05, min_samples=2
+    )
+    selected = selector(
+        candidates,
+        acquisition=_mock_acquisition([1.0, 1.0, 1.0, 1.0]),
+        cost_fn=_cost_fn([1.0, 1.0, 1.0, 1.0]),
+        round_budget=100.0,
+    )
+    assert len(selected) <= 2
```
carriepl: How would len(selected) == 1 be a valid test result, here? From the parameters defined, it seems like the solution will necessarily have to select 2 candidates.

src/activelearning/selector/knapsack_selector.py
pedromsferraz-mila: The CBC is a popular open-source MILP solver, changed wording to make the param descriptions more clear: fc502e31e85ea0a303f41bbf55132b25f5e18d28

src/activelearning/selector/knapsack_selector.py
pedromsferraz-mila: Good point, added a discussion on when to consider using warm_start: bc78cc7e05e8c4abe4d04b3a5142604af0febc66

src/activelearning/selector/knapsack_selector.py
pedromsferraz-mila: Good point. I agree that this selector only makes sense for acquisition scores representing meaningful additive utilities (for example, selecting five items with utility 1 should be equivalent to selecting one item with utility 5). This holds approximately for MF-MES when assuming uncorrelated items, but not for ordinal or non-additive scores, such as log-probabilities.

I therefore updated both `KnapsackSelector` and `ClusteredKnapsackSelector` to document this contract and raise a descriptive error for negative acquisition scores rather than silently clipping or shifting them. I also added tests covering this validation for both selectors.

Let me know what you think! 63f409fed066cc7d657cc748b30908f6d538bfb8

pedromsferraz-mila: Due to numerical or finite-sample estimation error, theoretically non-negative acquisition functions such as MF-MES and EI can occasionally produce small negative values. I now clamp these values to zero within the corresponding acquisition implementations, preserving their theoretical values and ensuring compatibility with the knapsack selectors: 85072e0df9ef507bff040c036c7979a600f2ee65

src/activelearning/selector/knapsack_selector.py
pedromsferraz-mila: It does, but PuLP returns variable values as float, and CBC's internal tolerances mean a "selected" binary can come back as something like 0.9999998 rather than a clean 1.0. Thresholding at 0.5 is the standard way to handle this robustly. Added a comment to clarify: d55fcf545aa3f922e769eceedecfba66c43f30f4

src/activelearning/selector/knapsack_selector.py
pedromsferraz-mila: Good point, changed MIP/MILP terminology in favour of plain language descriptions: 4bf34438060d8aedf8e6d742d00d4f3d0dea2c65

src/activelearning/selector/knapsack_selector.py
pedromsferraz-mila: Good point, added an example referencing `ClusteredKnapsackSelector`: fd252b9816146ba926fff6a66b484e3296c0818e

tests/selector/test_knapsack_selector.py
pedromsferraz-mila: Good point, added: 512c1d4f0f4e8139991f24cd62dfbecaaa26c092

src/activelearning/selector/knapsack_selector.py
pedromsferraz-mila: Good point. I don’t think the selector should raise an exception here: selecting an empty set is a valid knapsack solution when no candidate is affordable. The framework already handles this by terminating cleanly when the selector returns no samples.

For stochastic, candidate-dependent costs, perhaps a useful follow-up on a separate PR could be to add an optional bounded retry mechanism before terminating. What do you think?

tests/selector/test_knapsack_selector.py
pedromsferraz-mila: Yes, providing an early feasible incumbent is a valid reason to recommend `warm_start`, particularly when `time_limit`  is set. The greedy solution gives CBC a candidate solution at the beginning of the search, so if the time limit is reached, CBC may be able to return that incumbent rather than ending without one.

Note that this is not a guarantee: CBC must accept the supplied MIP start, and it must satisfy every model constraint, including any additional constraints added by subclasses. `warm_start` also does not help infeasible models.

Without a time limit, CBC normally continues until it proves optimality or determines that the model cannot be solved, so the warm start is less valuable and adds the cost of computing and loading the greedy solution. It can also occasionally slow solving because the initial incumbent changes branch-and-bound exploration.

I've added more details in the docstrings and a warning if warm starting is enabled without a time limit: 1be562ee12ceb728b2f75f5454ed8ca39e92a78b

Please let me know what you think!

src/activelearning/selector/knapsack_selector.py
pedromsferraz-mila: Update: a later comment prompted more changes related to warm starts 1be562ee12ceb728b2f75f5454ed8ca39e92a78b

tests/selector/test_knapsack_selector.py
pedromsferraz-mila: Good point, added: f82fa27b0b9890d2d02a1b35f6f59e585e4880db

tests/selector/test_clustered_knapsack_selector.py
pedromsferraz-mila: Good catch, this should be == 2, fixed: 1ba7b7c77f9f33213c999a41a596b18006e01373

tests/selector/test_clustered_knapsack_selector.py
pedromsferraz-mila: Good catch, fixed: fea083d6220fdfc79560ec0c4c7cd48b21209d55

PR #7, opened 2026-03-23
Singleton Acquisition Functions API & BoTorch Wrappers

olutayo-mila:
## Summary

This PR introduces an extensible API for **singleton (independent) acquisition functions** built on top of BoTorch.

It adds:
- BoTorch-backed acquisition wrappers (analytic + multi-fidelity)
- Flexible candidate set construction
- Optional cost-aware and multi-fidelity support
- Surrogate utilities to support acquisition workflows

⚠️ Scope: **singleton scoring only** (each candidate evaluated independently).  
Batch (q > 1) support will follow in a separate PR.

---

### 1. Core Acquisition Framework

- Updated `Acquisition` ABC (`acquisition.py`):
  - Explicit support for singleton vs batch scoring
  - Capability flags:
    - `supports_singleton_scoring`
    - `supports_batch_scoring`
  

- New BoTorch base classes (`botorch_acquisition.py`):
  - `BoTorchAcquisitionBase`  
    Handles surrogate validation, lifecycle, multi-fidelity helpers, and optional cost wiring

  - `AnalyticBoTorchAcquisition`  
    For exact (non-Monte Carlo (MC)) acquisitions → singleton scoring

  - `QBatchBoTorchAcquisition`  
    For MC/q-batch acquisitions → currently evaluated as q=1

---

#### Analytic (Exact) (`botorch_analytic.py`)
- UCB, EI, LogEI, PI, LogPI, PosteriorMean

#### Multi-Fidelity (MC-based) (`botorch_multifidelity.py`)
- qMFMES
- qMFLBMES (GIBBON-style lower bound)
- qMFKG

Features:
- Automatic target-fidelity resolution from surrogate
- Configurable candidate sets for entropy-based methods
- Optional cost-aware utility integration
- Optional post-hoc cost weighting via `score()`

---

### 3. Candidate Set API (`candidate_set.py`)

Flexible construction of discrete candidate sets:
- `HypercubeCandidateSetSpec` (uniform / LHS sampling)
- `TrainDataCandidateSetSpec` (reuse observed data)
- `TensorCandidateSetSpec` (user-provided)

---

### 4. Surrogate Enhancements

New public methods in `BoTorchGPSurrogate` (`botorch_surrogate.py`):
- Encoding:
  - `encode_candidates`
  - `encode_candidate_batches`
- Data access:
  - `get_train_data`
- Multi-fidelity utilities:
  - `get_fidelity_dimension`
  - `get_target_fidelity_value`
  - `is_multi_fidelity`

---

### Scoring Scope
- Only **singleton scoring** is supported (candidates evaluated independently)
- Analytic acquisitions are inherently singleton
- Multi-fidelity (MC / q-batch) acquisitions are evaluated in **singleton mode** (q = 1)
- **Joint batch scoring (q > 1)** is not yet supported

### Cost Handling
- **Integrated**: via BoTorch `cost_aware_utility`
- **Post-processing**: via `cost_weighting` in `score()`

### Configuration
- Candidate sets are explicitly defined via `CandidateSetSpec`
- Multi-fidelity settings default to surrogate metadata but can be overridden

---

### Code
- Added BoTorch acquisition base classes
- Added analytic + multi-fidelity acquisition wrappers
- Added candidate set specification system
- Extended surrogate API for acquisition compatibility
- Updated `Acquisition` interface

### Tests
- Added tests for each new functionality (see `test_botorch_acquisition.py`, `test_candidate_set.py`)

---

## Example

```python
# Fit surrogate
surrogate.fit(observations)

# Analytic acquisition
ucb = UpperConfidenceBound(beta=2.0)
ucb.update(surrogate, observations)
scores = ucb.score(candidates)

# Multi-fidelity acquisition
qmfmes = QMultiFidelityMaxValueEntropy(candidate_set_spec=spec)
qmfmes.update(surrogate, observations)
scores = qmfmes.score(candidates)

# Optional cost weighting
scores = qmfmes.score(candidates, cost_weighting=my_cost_fn)

### CODE REVIEW

File: src/activelearning/selector/score_selector.py [hunk 1]
```diff
@@ -43,7 +43,14 @@ def __call__(
         """
         if acquisition is None:
             raise ValueError("Acquisition function is required for ScoreSelector.")
```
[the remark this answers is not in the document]
olutayo-mila: Beyond the scope of this PR.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
```diff
@@ -0,0 +1,291 @@
...
+        expand: Optional[Callable[[torch.Tensor], torch.Tensor]] = None,
+        **kwargs: Any,
+    ) -> None:
+        if num_fantasies <= 0:
+            raise ValueError(f"num_fantasies must be > 0, got {num_fantasies}")
+        if num_mv_samples <= 0:
+            raise ValueError(f"num_mv_samples must be > 0, got {num_mv_samples}")
+        if num_y_samples <= 0:
+            raise ValueError(f"num_y_samples must be > 0, got {num_y_samples}")
+
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
```diff
@@ -0,0 +1,291 @@
...
+        self,
+        *,
+        num_fantasies: int = 64,
+        current_value: Optional[float] = None,
+        expand: Optional[Callable[[torch.Tensor], torch.Tensor]] = None,
+        **kwargs: Any,
+    ) -> None:
+        if num_fantasies <= 0:
+            raise ValueError(f"num_fantasies must be > 0, got {num_fantasies}")
+
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,520 @@
...
+    - validation and storage of a typed ``BoTorchGPSurrogate``,
+    - lifecycle management of the internal BoTorch acquisition object,
+    - multi-fidelity metadata resolution,
+    - optional target-fidelity projection plumbing,
+    - optional cost-aware utility plumbing.
+
+    Concrete subclasses are responsible for constructing the actual BoTorch
+    acquisition function by implementing ``_build_botorch_acquisition()``.
+    Scoring mechanics are intentionally left to intermediate subclasses such as
+    analytic or q-batch acquisition families.
```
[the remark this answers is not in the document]
olutayo-mila: Good point. Fixed a44e1701c44215dba86e60ef566e863e1a595a36.

tests/acquisition/test_botorch_acquisition.py
```diff
@@ -0,0 +1,765 @@
...
+        self,
+        fitted_surrogate: BoTorchGPSurrogate,
+        single_fidelity_observations: list[Observation],
+    ) -> None:
+        acq = StubAnalytic()
+        acq.update(fitted_surrogate, single_fidelity_observations)
+        first_acqf = acq.botorch_acqf
+
+        acq.update(fitted_surrogate, single_fidelity_observations)
+        assert acq.botorch_acqf is not first_acqf or first_acqf is not None
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed 905818a40e0fbba0e1332621295c34198f9fce6d.

tests/acquisition/test_botorch_acquisition.py
```diff
@@ -0,0 +1,765 @@
...
+        single_fidelity_observations: list[Observation],
+        candidates: list[Candidate],
+    ) -> None:
+        """Scoring should not accumulate gradients."""
+        acq = StubAnalytic()
+        acq.update(fitted_surrogate, single_fidelity_observations)
+        acq.score(candidates)
+        # If torch.no_grad was not applied, the acqf output would track grad.
+        # We verify indirectly: no gradient-related errors were raised.
+
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed 5c3e28cd5e25ad1c295df1809c57796b71e3b8fa.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,515 @@
...
+        TypeError
+            If the surrogate is not a ``BoTorchGPSurrogate``.
+        """
+        super().update(surrogate, observations)
+
+        if not isinstance(surrogate, BoTorchGPSurrogate):
+            raise TypeError(
+                f"{self.__class__.__name__} requires a BoTorchGPSurrogate, "
+                f"but received {type(surrogate).__name__}."
+            )
```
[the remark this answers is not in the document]
olutayo-mila: Good point. Fixed bcb9d1fcb58d5e31afe40a324b8178fc2a157505.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
```diff
@@ -0,0 +1,291 @@
...
+        current_value = (
+            torch.tensor(self._current_value)
+            if self._current_value is not None
+            else None
+        )
+
+        build_kwargs: dict[str, Any] = {
+            "model": self._botorch_surrogate.get_model(),
+            "num_fantasies": self._num_fantasies,
+            "current_value": current_value,
```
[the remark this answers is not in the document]
olutayo-mila: Great point! Thanks. Fixed fb47fd9c318979282c70e03412269e7800212c2f.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
```diff
@@ -0,0 +1,306 @@
...
+        model.eval()
+        with torch.no_grad():
+            model.posterior(train_X)
+
+        current_value = (
+            torch.tensor(self._current_value, dtype=torch.float64)
+            if self._current_value is not None
+            else None
+        )
+
```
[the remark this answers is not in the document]
olutayo-mila: Thanks. Fixed fb47fd9c318979282c70e03412269e7800212c2f.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
```diff
@@ -0,0 +1,307 @@
...
+        build_kwargs: dict[str, Any] = {
+            "model": model,
+            "num_fantasies": self._num_fantasies,
+            "current_value": current_value,
+        }
+
+        if not self.maximize:
+            build_kwargs["posterior_transform"] = ScalarizedPosteriorTransform(
+                weights=torch.tensor([-1.0], dtype=torch.float64)
+            )
```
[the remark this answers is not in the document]
olutayo-mila: Good point. Fixed 742a0b63d9ea15f191a76f8bb64c3bfa858d26f1.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,505 @@
...
+        """
+        if best_f_override is not None:
+            return best_f_override
+
+        if self._observations_cache is not None and len(self._observations_cache) > 0:
+            ys = [float(obs.y) for obs in self._observations_cache]
+            return max(ys) if self.maximize else min(ys)
+
+        if self._botorch_surrogate is not None:
+            _, train_Y = self._botorch_surrogate.get_train_data()
```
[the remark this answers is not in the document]
olutayo-mila: Fair point. Fixed 742a0b63d9ea15f191a76f8bb64c3bfa858d26f1.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,510 @@
...
+        return self._resolved_cost_model
+
+    @property
+    def resolved_cost_aware_utility(self) -> Optional[Any]:
+        """Resolved cost-aware utility, or ``None`` if not used."""
+        return self._resolved_cost_aware_utility
+
+    def update(
+        self,
+        surrogate: Surrogate,
```
pedromsferraz-mila, line 91: As you expect a `BoTorchGPSurrogate` here, you can change the type annotation from `Surrogate` to `BoTorchGPSurrogate`.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,510 @@
...
+    ``_build_botorch_acquisition()`` to construct the BoTorch acquisition
+    and optionally ``_score_encoded()`` to customize how scores are computed.
+    """
+
+    def __init__(
+        self,
+        *,
+        maximize: bool = True,
+        target_fidelity_value: Optional[float] = None,
+        project_to_target_fidelity_fn: Optional[
```
pedromsferraz-mila, line 35: From the variable names and the docstring, `target_fidelity_value` and `project_to_target_fidelity_fn` are still not clear enough to me. I suggest renaming them to be more explicit and better aligned with the terminology used in Alex’s paper (e.g., `highest_fidelity_value`).

Also, my understanding is that `project_to_target_fidelity_fn` simply sets the fidelity level to the highest fidelity, is that correct? If so, I’d recommend simpler naming (e.g., `set_to_highest_fidelity_fn`), since “project” suggests a more complex operation than what this function appears to do.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,510 @@
...
+
+    @property
+    def resolved_cost_model(self) -> Optional[Any]:
+        """Resolved cost model, or ``None`` if not used."""
+        return self._resolved_cost_model
+
+    @property
+    def resolved_cost_aware_utility(self) -> Optional[Any]:
+        """Resolved cost-aware utility, or ``None`` if not used."""
+        return self._resolved_cost_aware_utility
```
pedromsferraz-mila: What do you think about removing the public properties here? Looking at the codebase, they're only ever used in tests, no external module actually depends on them. Since they mostly expose internal resolved state rather than anything that should be part of the public API, I feel like removing them would reduce cognitive load and keep the class simpler overall.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,510 @@
...
+        self,
+        *,
+        maximize: bool = True,
+        target_fidelity_value: Optional[float] = None,
+        project_to_target_fidelity_fn: Optional[
+            Callable[[torch.Tensor], torch.Tensor]
+        ] = None,
+        fidelity_costs: Optional[dict[int, float]] = None,
+        cost_model: Optional[Any] = None,
+        cost_aware_utility: Optional[Any] = None,
```
pedromsferraz-mila: Do these parameters interact with each other? If so, what's the relationship between them? Could be worth describing in the docstring.

pedromsferraz-mila: Also, if there are conflicts, it would be nice to either to log a warning message or raise an error.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,510 @@
...
+
+        Returns
+        -------
+        result : Any
+            Concrete BoTorch acquisition object built from the current surrogate
+            state, observations cache, and any resolved MF / cost-aware helpers.
+        """
+        pass
+
+    def _resolve_best_f(self, best_f_override: Optional[float] = None) -> float:
```
pedromsferraz-mila, line 259: Looking at the codebase, this method is currently only used by analytic acquisition functions. If that's true in general, I'd suggest moving it to `AnalyticBoTorchAcquisition`.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+        if self._cost_aware_utility_override is not None:
+            return self._cost_aware_utility_override
+
+        if self._resolved_cost_model is None:
+            return None
+
+        return self._build_default_cost_aware_utility(self._resolved_cost_model)
+
+    def _build_default_cost_model_from_fidelity_costs(
```
pedromsferraz-mila: Currently, if a user passes `fidelity_costs`, it causes an unexpected `NotImplementedError` during  `update()` because this method isn't implemented in a subclass yet. In this context, I'd suggest removing  `fidelity_costs` and this method for now until the feature is actually implemented to make the API as clean as possible.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,510 @@
...
+        ------
+        NotImplementedError
+            Default fidelity-cost-based cost modeling has not yet been
+            implemented in the base class.
+        """
+        raise NotImplementedError(
+            "Default cost-model construction from fidelity costs is not implemented yet."
+        )
+
+    def _build_default_cost_aware_utility(self, cost_model: Any) -> Any:
```
pedromsferraz-mila: Similarly to above, as this method is still not implemented by any subclass, I'd suggest removing it until it's implemented to simplify the API.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
```diff
@@ -0,0 +1,309 @@
...
+        super().__init__(**kwargs)
+        self._candidate_set_spec = candidate_set_spec
+        self._num_fantasies = num_fantasies
+        self._num_mv_samples = num_mv_samples
+        self._num_y_samples = num_y_samples
+        self._expand = expand
+
+    def update(
+        self,
+        surrogate: Surrogate,
```
pedromsferraz-mila, line 94: Similarly to the base class, consider changing the type annotation to `BoTorchGPSurrogate`.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,510 @@
...
+            ``BoTorchGPSurrogate``.
+        observations : Optional[Iterable[Observation]]
+            Optional iterable of current observations.
+
+        Raises
+        ------
+        TypeError
+            If the surrogate is not a ``BoTorchGPSurrogate``.
+        """
+        super().update(surrogate, observations)
```
pedromsferraz-mila: There's a small subtlety here: if `super().update(surrogate, observations)` is called with the raw iterable, then `list(observations)` is called on line 151. If the base ever iterates the observations, `list(observations)` would silently return an empty list. Materializing first and passing the list to both is safer:
```
   obs_list = list(observations) if observations is not None else None
   super().update(surrogate, obs_list)
   ...
   self._observations_cache = obs_list
```

src/activelearning/acquisition/botorch/botorch_multifidelity.py
```diff
@@ -0,0 +1,309 @@
...
+            build_kwargs["cost_aware_utility"] = self._resolved_cost_aware_utility
+        if self._resolved_project_to_target_fidelity_fn is not None:
+            build_kwargs["project"] = self._resolved_project_to_target_fidelity_fn
+        if self._expand is not None:
+            build_kwargs["expand"] = self._expand
+
+        return _qMFMES(**build_kwargs)
+
+
+class QMultiFidelityLowerBoundMaxValueEntropy(QBatchBoTorchAcquisition):
```
pedromsferraz-mila: Currently, the `QMultiFidelityMaxValueEntropy` and` 
  QMultiFidelityLowerBoundMaxValueEntropy` classes are very similar (they share identical `__init__` and `update()` and almost identical `_build_botorch_acquisition()`). Consider adding an intermediate base (e.g. `_QMultiFidelityEntropyBase`) with the shared logic and subclasses only overriding `_build_botorch_acquisition`.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
```diff
@@ -0,0 +1,309 @@
...
+            Current observations forwarded to the candidate set spec and base.
+        """
+        if observations is not None:
+            obs_list = list(observations)
+            self._candidate_set_spec.update(obs_list)
+            super().update(surrogate, obs_list)
+        else:
+            super().update(surrogate, observations)
+
+    def _build_botorch_acquisition(self) -> Any:
```
pedromsferraz-mila, line 117: Missing docstring (for consistency with other subclasses).

src/activelearning/acquisition/botorch/candidate_set.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+class CandidateSetSpec(ABC):
+    """Base class for candidate-set construction strategies.
+
+    A :class:`CandidateSetSpec` defines how to produce the discrete set of
+    points used by entropy-based acquisition functions to approximate the
+    distribution of the optimum.
+
+    The returned tensor is in **model space** — the feature representation
```
pedromsferraz-mila, line 46: This class (and subclasses) expects the data to be formatted in the BoTorch format (i.e., torch tensors with dimensions (N, d) including the fidelity). 

In this context, I think it'd be more clear to explicitly state this in the docstring and update the type annotations to reflect that (e.g. using `BoTorchGPSurrogate` instead of `Surrogate`). 

In that case, I'd also suggest moving this file into the `src/activelearning/acquisition/botorch` subfolder.

Let me know what you think!

src/activelearning/acquisition/botorch/candidate_set.py
```diff
@@ -0,0 +1,302 @@
...
+    sampling procedure.
+
+    Parameters
+    ----------
+    tensor : torch.Tensor
+        Candidate set in model space.  Shape ``(N, d)``.
+    """
+
+    def __init__(self, tensor: torch.Tensor) -> None:
+        self.tensor = tensor
```
pedromsferraz-mila, line 293: Here it'd be nice to add some checks on the tensor dimensions to ensure they're compatible with the expected shape (e.g. asserting `ndim == 2`).

src/activelearning/sampler/pool_score_sampler.py
```diff
@@ -49,8 +49,12 @@ def sample(
 
         if acquisition is None:
             raise ValueError("Acquisition function is required for PoolScoreSampler.")
+        if not acquisition.supports_singleton_scoring:
```
pedromsferraz-mila: I believe we don't need these checks before calling the `acquisition.score()` method if we're throwing an error anyway. Since `Acquisition.score()` already raises a `NotImplementedError` with an informative message if the method isn't implemented (defined in the base `Acquisition` class, see lines 126–128 of `acquisition.py`), this ends up being redundant.

The `supports_singleton_scoring` guard is really only useful if the caller wants to branch (e.g., to fall back to a different strategy), but here it just gates an error with another error. Removing it keeps the code DRY without sacrificing clarity.

src/activelearning/selector/cost_aware_selector.py
```diff
@@ -56,9 +56,14 @@ def __call__(
 
         if not candidates:
             return []
+        if not acquisition.supports_singleton_scoring:
```
pedromsferraz-mila: Same comment as above on unnecessary check.

src/activelearning/selector/score_selector.py
```diff
@@ -43,7 +43,14 @@ def __call__(
...
-
-        acq_values = acquisition(candidates)
+        if not candidates:
+            return []
+        if not acquisition.supports_singleton_scoring:
```
pedromsferraz-mila: Same comment as above on unnecessary check.

pedromsferraz-mila (commented):
Great work, this provides plenty of options for acquisition functions! Most of my comments are about cleaning up unnecessary code to keep the API as lean and easy to read as possible.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: I agree. Fixed 14d1fe73f4833b0c39a718e31eb0260f04c73318.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Good point. It is only used by analytic acquisitions in this PR. However, there are other qbatch acquisitions that also use this method (e.g., qEI, qPI) so I’d prefer to keep it in the common base class.

pedromsferraz-mila: Thanks for the clarification, that sounds good to me!

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Good point, and I agree that in most practical cases the projection will indeed be to the highest fidelity. However, I think we should still keep the “target fidelity” / “project” terminology to stay aligned with BoTorch’s API and also to keep the abstraction slightly more general. In principle, the projection is defined with respect to a chosen target fidelity, even if in practice that is typically the highest level. Also, since we allow a custom projection callable, the operation is not strictly limited to setting a single value. 

I’ve updated the docstrings to make this clearer as well. See here 506fd6c4f273362974676bf9287da8ba5a050d09.

Let me know what you think, and happy to adjust if needed. Thanks!

pedromsferraz-mila: I see, thanks for the clarification! That looks good to me!

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Good point, thanks! Yes, these parameters are related. The cost_aware_utility typically wraps a BoTorch cost model, so specifying both separately is redundant. In practice, if a user provides a cost_aware_utility, it already encapsulates the cost model, so a separate cost_model is actually not needed. I’ve updated the code and docstring accordingly. See here 538d1237e4c19bdffb1fde77455ce1bb395a70b3.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Good point, thanks! Fixed 7527134b2c63c36e0a85ccb0531b643fb32c4b09.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Good point! Fixed in the same commit as above: 7527134b2c63c36e0a85ccb0531b643fb32c4b09.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Thanks! Good catch! Fixed ef4df5da089a5a16898d7fc8b8c9621b214ac2c6.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
olutayo-mila: Thanks, I agree. Fixed 165feeb58121d6ece26d446d20605257129ebae3.

PS: Minor edit to the commit above here 1aaeaeafed474395e9cfebfe6bd694a2732af7c7.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
olutayo-mila: Thanks. No longer necessary due to the above refactoring.

src/activelearning/acquisition/botorch/candidate_set.py
olutayo-mila: Thanks! I agree. Fixed 19c89153c0c9bd31a48f0b65727566d82c0b2383.

src/activelearning/acquisition/botorch/candidate_set.py
olutayo-mila: Thanks. I agree. Fixed 2480c7138636b889ccdc8ae06c3689604f929c40.

src/activelearning/sampler/pool_score_sampler.py
olutayo-mila: Fair point. Fixed 2d1bde8e66d3025b34fcf99c173afe5a302b8eab.

src/activelearning/selector/cost_aware_selector.py
olutayo-mila: Fixed 2d1bde8e66d3025b34fcf99c173afe5a302b8eab.

src/activelearning/selector/score_selector.py
olutayo-mila: Fixed 2d1bde8e66d3025b34fcf99c173afe5a302b8eab.

src/activelearning/acquisition/botorch/botorch_acquisition.py
pedromsferraz-mila: Perfect, that looks great, thanks!

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Thanks for the suggestion. I did consider narrowing the annotation to BoTorchGPSurrogate, but that causes an LSP/type-checking issue since Acquisition.update() is defined with surrogate: Surrogate. Narrowing the parameter type in the subclass breaks substitutability.

For that reason, I kept the annotation as Surrogate and enforce the stronger requirement with the runtime isinstance(..., BoTorchGPSurrogate) check.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
olutayo-mila: Same type check issue as in the base class.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
pedromsferraz-mila: That looks great, thanks!

src/activelearning/acquisition/botorch/botorch_acquisition.py
pedromsferraz-mila: Ahh yes, good point! That looks good to me, thank you!

src/activelearning/acquisition/acquisition.py
```diff
@@ -1,79 +1,198 @@
...
+        candidates: Iterable[Candidate],
+        cost_weighting: Optional[
+            Callable[[list[float], list[Candidate]], list[float]]
+        ] = None,
+    ) -> list[float]:
+        """Score candidates independently.
+
+        This method is intentionally optional. Subclasses that only support
+        joint batch scoring need not implement it. Callers should check
+        ``supports_singleton_scoring`` before calling this method.
```
alexhernandezgarcia, line 99: Question: I don't understand why an acquisition subclass would not support independent scoring? Isn't that always possible?

src/activelearning/acquisition/dummy_acquisition.py
```diff
@@ -44,11 +50,12 @@ def __call__(self, candidates: Sequence[Candidate]) -> list[float]:
         ValueError
             If predict() does not return required "mean" key.
         """
+        candidate_list = list(candidates)
```
alexhernandezgarcia: It would have relatively small impact but I would suggest the following:
- If `self.surrogate is None` then `candidate_list` isn't really needed in line 55 (only the number of candidates), so there is no need to always materialise it before.
- Then in line 58 I would do `pred = self.surrogate.predict(list(candidates))`

alexhernandezgarcia: Oh I have just seen that `candidate_list` is also used for cost weighting later. So the last point would not be correct. Still, `candidate_list = list(candidates)` could be done after line 55.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,396 @@
...
+    ``_build_botorch_acquisition()`` to construct the BoTorch acquisition
+    and optionally ``_score_encoded()`` to customize how scores are computed.
+    """
+
+    # Overridden to True by multi-fidelity subclasses.
+    _supports_multi_fidelity: ClassVar[bool] = False
+
+    def __init__(
+        self,
+        *,
```
alexhernandezgarcia, line 32: Excuse my ignorance, but what is this for?

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,396 @@
...
+
+        # Resolved helpers populated during update()
+        self._resolved_target_fidelity_value: Optional[float] = None
+        self._resolved_project_to_target_fidelity_fn: Optional[
+            Callable[[torch.Tensor], torch.Tensor]
+        ] = None
+
+    def update(
+        self,
+        surrogate: Surrogate,
```
alexhernandezgarcia, line 91: Would it make sense to change the type to `BoTorchGPSurrogate`?

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,396 @@
...
+
+        Parameters
+        ----------
+        surrogate : Surrogate
+            Surrogate model to use for subsequent acquisition scoring. Must be a
+            ``BoTorchGPSurrogate``.
+        observations : Optional[Iterable[Observation]]
+            Optional iterable of current observations.
+
+        Raises
```
alexhernandezgarcia, line 108: Add the multi-fidelity warning too?

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,396 @@
...
+            If projection is requested in multi-fidelity mode but required
+            fidelity metadata is unavailable.
+        """
+        if self._project_to_target_fidelity_fn_override is not None:
+            return self._project_to_target_fidelity_fn_override
+
+        if self._botorch_surrogate is None:
+            return None
+        if not self._botorch_surrogate.is_multi_fidelity():
+            return None
```
alexhernandezgarcia: Would it make sense to move this to the top? This would explicitly indicate that this is relevant only if the surrogate is multi-fidelity, which is the case, right?

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -0,0 +1,396 @@
...
+        if self._botorch_acqf is None:
+            return [1.0] * len(cand_list)
+        X = self._botorch_surrogate.encode_candidates(cand_list).unsqueeze(1)  # type: ignore[union-attr]
+        raw_scores = self._score_encoded(X)
+        if cost_weighting is None:
+            return raw_scores
+        return cost_weighting(raw_scores, cand_list)
+
+
+class AnalyticBoTorchAcquisition(BoTorchAcquisitionBase):
```
alexhernandezgarcia, line 362: What is the purpose of AnalyticBoTorchAcquisition and QBatchBoTorchAcquisition if they are the same as BoTorchAcquisitionBase? (I'm most likely missing something, sorry)

src/activelearning/acquisition/acquisition.py
olutayo-mila: Good question. It’s not always possible in a meaningful way because while singleton scoring can sometimes be viewed as the q=1 case of batch scoring, that depends on the acquisition being well-defined for singletons.

Since this is an abstract base class, subclasses may define acquisitions that are inherently joint or only implemented for batch inputs. So we allow them to explicitly support either mode rather than assuming singleton scoring always exists.

Curious to hear your thoughts.

src/activelearning/acquisition/dummy_acquisition.py
olutayo-mila: Good catch! Fixed e9fe3ff83f4b3f700bc8fdf4bd11be521a62f527.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: No worries. The '*' forces all arguments to be keyword-only. It's used here because the BoTorch constructors have several optional parameters where positional calls could be ambiguous.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Thanks for raising this. I did try narrowing the type to BoTorchGPSurrogate, but it causes issues with the type checker. The base Acquisition class accepts a Surrogate, and narrowing the parameter type in a subclass would violate the Liskov Substitution Principle i.e., the subclass would no longer be substitutable wherever an Acquisition is expected.

So I kept the more general type and instead validate the specific surrogate type at runtime where needed.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Good catch! Fixed 0d340369dca3e3d7cf58ee8cf9e6f5d6acc984a1.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Good catch, thanks! Fixed 6e6640ee36bb285cc0a376e76788a7edc0eaf241.

src/activelearning/acquisition/botorch/botorch_acquisition.py
olutayo-mila: Good question. They’re identical right now, but intentionally so.

The goal is to separate analytic vs q-batch BoTorch acquisitions at the type level, so concrete subclasses inherit from the correct branch. This keeps the hierarchy clean and makes it easier to introduce category-specific behavior later without mixing concerns.

Happy to revisit if this feels premature.

src/activelearning/acquisition/botorch/botorch_acquisition.py
alexhernandezgarcia: Thanks for explaining! Nice to learn something new :)

src/activelearning/acquisition/botorch/botorch_acquisition.py
alexhernandezgarcia: Thanks for explaining and for trying. No big deal to keep it more general.

src/activelearning/acquisition/botorch/botorch_acquisition.py
alexhernandezgarcia: That's great forward thinking, I love it! :) Thanks!

src/activelearning/acquisition/botorch/botorch_multifidelity.py
alexhernandezgarcia, line 1: I was thinking that it would be nice to include some references to the botorch tutorials and/or papers relevant to each acquisition function, for example:
- https://botorch.org/docs/tutorials/max_value_entropy/
- https://arxiv.org/abs/1901.08275

This is just a nice thing to have, completely optional. I could take care of this later on.

src/activelearning/acquisition/botorch/candidate_set.py
```diff
@@ -0,0 +1,313 @@
+"""Candidate-set specifications for entropy-based acquisition functions.
+
+Some acquisition functions — such as Max-Value Entropy Search (MES) variants —
+require a discrete set of candidate points to approximate the distribution of
+the optimum.  The quality of this approximation depends on how well the
```
alexhernandezgarcia: Extra space
```suggestion
the optimum. The quality of this approximation depends on how well the
```

src/activelearning/acquisition/botorch/candidate_set.py
```diff
@@ -0,0 +1,313 @@
...
+    distribution of the optimum.
+
+    The returned tensor is in **model space** — the feature representation
+    expected by the surrogate's internal model — with shape ``(N, d)``, where
+    ``d`` includes any fidelity column in multi-fidelity settings.
+
+    Implement this class to define custom candidate-set strategies.
+    """
+
+    def update(self, observations: Iterable[Observation]) -> None:
```
alexhernandezgarcia, line 53: Add `return None`?

(as in https://github.com/milaforscience/activelearning/blob/main/src/activelearning/surrogate/surrogate.py#L125)

src/activelearning/acquisition/botorch/candidate_set.py
```diff
@@ -0,0 +1,313 @@
...
+    # ------------------------------------------------------------------
+    # Private helpers
+    # ------------------------------------------------------------------
+
+    def _sample_unit(self) -> torch.Tensor:
+        """Return ``(n_points, n_dims)`` samples in ``[0, 1]^d``."""
+        n_dims = len(self.bounds)
+        if self.strategy == "lhs":
+            return latin_hypercube(self.n_points, n_dims)
+        return torch.rand(self.n_points, n_dims, dtype=torch.float64)
```
alexhernandezgarcia: Consider using `dtype=self.dtype` from the runtime context, introduced in PR https://github.com/milaforscience/activelearning/pull/5/.

alexhernandezgarcia: Same would go to other cases in this PR. I will not comment in the other cases so as not to clutter the PR with comments.

src/activelearning/acquisition/botorch/candidate_set.py
```diff
@@ -0,0 +1,313 @@
...
+            return torch.cat([feature_points, fid_col], dim=-1)
+
+        return feature_points
+
+
+class TrainDataCandidateSetSpec(CandidateSetSpec):
+    """Build a candidate set from observed data.
+
+    The candidate set is built from the candidates seen in the observations
+    provided via :meth:`update`.  This is a convenient default for
```
alexhernandezgarcia, line 203: Extra space
```suggestion
    provided via :meth:`update`. This is a convenient default for
```

src/activelearning/acquisition/botorch/candidate_set.py
```diff
@@ -0,0 +1,313 @@
...
+        candidate_set : torch.Tensor
+            BoTorch-ready candidate tensor of shape ``(N, d)``.
+        """
+
+
+class HypercubeCandidateSetSpec(CandidateSetSpec):
+    """Build a candidate set by sampling from a bounded hypercube.
+
+    Generates points by uniform or Latin Hypercube Sampling within the
+    specified feature bounds.  Suitable for **continuous** domains.
```
alexhernandezgarcia: Extra space
```suggestion
    specified feature bounds. Suitable for **continuous** domains.
```

src/activelearning/acquisition/botorch/candidate_set.py
```diff
@@ -0,0 +1,313 @@
...
+
+class TrainDataCandidateSetSpec(CandidateSetSpec):
+    """Build a candidate set from observed data.
+
+    The candidate set is built from the candidates seen in the observations
+    provided via :meth:`update`.  This is a convenient default for
+    **discrete** domains where the observed inputs form a natural support set.
+
+    Call :meth:`update` with the current observations before calling
+    :meth:`build`.  In the library, acquisitions that hold a
```
alexhernandezgarcia: Extra space
```suggestion
    :meth:`build`. In the library, acquisitions that hold a
```

src/activelearning/surrogate/botorch_surrogate.py
```diff
@@ -434,6 +453,105 @@ def encode_candidates(self, candidates: Iterable[Candidate]) -> torch.Tensor:
...
+        if ragged_indices:
+            raise ValueError(
+                "All candidate batches must have the same size. "
+                f"Ragged batch indices: {ragged_indices}."
+            )
+
+        encoded_batches = [self.encode_candidates(batch) for batch in batch_list]
+        return torch.stack(encoded_batches, dim=0)
+
+    def is_multi_fidelity(self) -> bool:
```
alexhernandezgarcia, line 499: A couple of comments on this:
- Would it make sense to make this method a `@property`?
- There are a bunch of places in this class and even in `botorch_acquisition.py` where `self._is_multi_fidelity` or `surrogate._is_multi_fidelity` is still used. Should those not be updated to `*.is_multi_fidelity()` or `*.is_multi_fidelity`  (if a property)?

src/activelearning/surrogate/surrogate.py
```diff
@@ -6,32 +6,28 @@
 
 
 class Surrogate(ABC, ALRuntimeMixin):
-    """Abstract surrogate interface used by acquisitions and the active learning loop.
+    """Abstract surrogate interface for approximating the objective function.
 
-    Surrogate models approximate the oracles based on observed data, enabling
-    efficient candidate evaluation.
+    Surrogate models approximate the objective based on observed data, enabling
+    efficient candidate evaluation without querying the oracle directly.
```
alexhernandezgarcia:
```suggestion
    efficient candidate evaluation without querying the oracles directly.
```

alexhernandezgarcia (dismissed):
Fantastic work! I have only dropped a few additional minor comments/suggestions. Overall, it looks good to me!

Note: I have not reviewed the tests.

Again, apologies for the delay in reviewing this PR.

src/activelearning/acquisition/botorch/candidate_set.py
olutayo-mila: Good catch! Fixed: 6170b8ed7d02d3ee0e621850c67f6ea539e7aa3e.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Yes, it would make more sense. Thanks for the suggestion. Fixed 91e26aedfa9f7bb22d070715aba904cbe046506b.

src/activelearning/acquisition/botorch/botorch_multifidelity.py
alexhernandezgarcia: I have set this as a [new issue](https://github.com/milaforscience/activelearning/issues/15). Marking this as resolved.

src/activelearning/acquisition/acquisition.py
alexhernandezgarcia: Thanks, that makes sense!

src/activelearning/surrogate/botorch_surrogate.py
alexhernandezgarcia: Thanks for making these changes!

Just one curiosity question: is it a deliberate decision that in `botorch_surrogate.py` still `self._is_multi_fidelity` is used instead of `self._is_multi_fidelity`?

alexhernandezgarcia (approved):
Re-approved! Thanks for the changes and answers to my questions.

src/activelearning/surrogate/botorch_surrogate.py
olutayo-mila: Yes, it's deliberate. The property `is_multi_fidelity` is read-only so external code cannot modify it. Internally, the class still uses `self._is_multi_fidelity` directly because it needs to assign  to it (e.g., during `fit()` and `update()`).

PR #10, opened 2026-04-02
Features/qbatch acquisitions

olutayo-mila:
# PR: q-batch acquisition functions and batch scoring

**Base branch:** `features/acquisition-api-and-botorch`  
**Head branch:** `features/qbatch-acquisitions`

---

## Summary

Extends the BoTorch acquisition layer introduced in the previous PR with nine
Monte Carlo (q-batch) acquisition functions and opt-in joint batch scoring.
This enables callers to evaluate a *set* of candidates jointly — the natural
mode for most BoTorch MC acquisitions — rather than independently.

---

### `BatchScoringMixin` (`botorch_acquisition.py`)

A lightweight mixin that adds `score_batches()` to any q-batch acquisition
that supports joint evaluation:

```python
class QExpectedImprovement(BatchScoringMixin, QBatchBoTorchAcquisition):
    ...
```

- `score_batches(candidate_batches, cost_weighting=None) → list[float]`
  encodes each batch as a `(q, d)` tensor and passes them as a single
  t-batched forward call `(B, q, d)`.
- Returns `[1.0] * B` before `update()` is called (consistent with
  `score()` behaviour on first round).
- `cost_weighting` post-processing follows the same pattern as `score()`.
- Acquisitions that do **not** mix in `BatchScoringMixin` retain
  `supports_batch_scoring = False` and raise `NotImplementedError` on
  `score_batches()` — explicit rather than silent.

### New file: `botorch_qbatch.py`

Eight concrete q-batch acquisition wrappers plus `QMaxValueEntropy`:

| Class | BoTorch backing | Batch scoring |
|---|---|---|
| `QExpectedImprovement` | `qExpectedImprovement` | ✅ |
| `QLogExpectedImprovement` | `qLogExpectedImprovement` | ✅ |
| `QNoisyExpectedImprovement` | `qNoisyExpectedImprovement` | ✅ |
| `QLogNoisyExpectedImprovement` | `qLogNoisyExpectedImprovement` | ✅ |
| `QUpperConfidenceBound` | `qUpperConfidenceBound` | ✅ |
| `QProbabilityOfImprovement` | `qProbabilityOfImprovement` | ✅ |
| `QSimpleRegret` | `qSimpleRegret` | ✅ |
| `QKnowledgeGradient` | `qKnowledgeGradient` | ❌ |
| `QMaxValueEntropy` | `qMaxValueEntropy` | ❌ |

Not all q-batch acquisitions support joint scoring. Some (such as
`QKnowledgeGradient` and `QMaxValueEntropy`) use one-shot or fantasy-based
optimisation internally, where jointly evaluating a pre-formed batch of
candidates does not produce meaningful scores. These omit `BatchScoringMixin`
and should be used via singleton `score()` instead.

### Minimisation support (`maximize=False`)

BoTorch MC acquisitions with the exception of qMES do not accept a `maximize` parameter directly (unlike
their analytic counterparts). When `maximize=False`, each wrapper applies a
`ScalarizedPosteriorTransform(weights=[-1])` to negate the posterior, and
reference values (`best_f`, `current_value`) are negated accordingly. This
ensures correct behaviour for minimisation objectives across all nine wrappers.

### Stale candidate-set warning

Entropy-based acquisitions that depend on a `CandidateSetSpec` (e.g.
`QMaxValueEntropy`) now emit a `UserWarning` when `update()` is called
without observations, alerting users that the candidate set may be stale.
The warning is scoped to specs that override `CandidateSetSpec.update()`,
so stateless specs like `HypercubeCandidateSetSpec` are not affected.

### Surrogate encoding optimisation

`encode_candidates()` and `encode_candidate_batches()` skip redundant
`list()` copies when the input is already materialised, avoiding
unnecessary allocations in the `score_batches()` path.

### Validation added

- `QKnowledgeGradient.__init__` raises `ValueError` if `num_fantasies <= 0`.
- `QMaxValueEntropy.__init__` raises `ValueError` if `num_fantasies`, `num_mv_samples`, or `num_y_samples` ≤ 0.

---

## Design decisions

**Why a mixin rather than a base class flag?**  
Joint batch scoring is an *opt-in capability*, not a default. Inheriting from
`BatchScoringMixin` is explicit at the class definition site and makes it
immediately clear which acquisitions support it. A boolean flag would require
reading the implementation to know whether `score_batches` is safe to call.

**Why `ScalarizedPosteriorTransform` for minimisation?**  
BoTorch's analytic acquisitions accept `maximize=True/False`, but the MC
variants do not. Negating the posterior via a `ScalarizedPosteriorTransform`
is the standard BoTorch pattern for minimisation in the MC setting. Reference
values (`best_f`, `current_value`) are negated to stay consistent with the
flipped objective space.

---

## Testing

- Integration tests for all nine acquisitions covering: surrogate update,
  valid float scores, constructor parameter forwarding.
- `maximize=False` tests for qEI, qNEI, qUCB, and qSR verifying
  scores differ from `maximize=True`.
- `BatchScoringMixin` tests: joint scoring after update, `cost_weighting`
  application, empty input handling, pre-update fallback to `1.0`.
- Stale candidate-set warning tests for both `QMaxValueEntropy` and
  `QMultiFidelityMaxValueEntropy`.
- Validation tests: `num_fantasies`/`num_mv_samples`/`num_y_samples` ≤ 0 raises `ValueError`.

**409 tests passing.**

### DIFF
--- a/src/activelearning/acquisition/botorch/botorch_qbatch.py
+++ b/src/activelearning/acquisition/botorch/botorch_qbatch.py
@@ -1,0 +1,479 @@
+"""Concrete wrappers for BoTorch q-batch (Monte Carlo) acquisition functions.
+
+Each class subclasses :class:`QBatchBoTorchAcquisition` and implements
+:meth:`_build_botorch_acquisition` to construct the corresponding BoTorch
+MC acquisition object.
+"""
+
+import warnings
+from typing import Any, Callable, Iterable, Optional
+
+import torch
+from botorch.acquisition.knowledge_gradient import (
+    qKnowledgeGradient as _qKG,
+)
+from botorch.acquisition.logei import (
+    qLogExpectedImprovement as _qLogEI,
+    qLogNoisyExpectedImprovement as _qLogNEI,
+)
+from botorch.acquisition.max_value_entropy_search import (
+    qMaxValueEntropy as _qMES,
+)
+from botorch.acquisition.monte_carlo import (
+    qExpectedImprovement as _qEI,
+    qNoisyExpectedImprovement as _qNEI,
+    qProbabilityOfImprovement as _qPI,
+    qSimpleRegret as _qSimpleRegret,
+    qUpperConfidenceBound as _qUCB,
+)
+from botorch.acquisition.objective import ScalarizedPosteriorTransform
+
+from activelearning.acquisition.botorch.botorch_acquisition import (
+    BatchScoringMixin,
+    QBatchBoTorchAcquisition,
+)
+from activelearning.acquisition.botorch.candidate_set import CandidateSetSpec
+from activelearning.surrogate.surrogate import Surrogate
+from activelearning.utils.types import Observation
+
+
+class QExpectedImprovement(BatchScoringMixin, QBatchBoTorchAcquisition):
+    """Monte Carlo q-Expected Improvement (qEI).
+
+    Parameters
+    ----------
+    best_f : float, optional
+        Best observed objective value. If ``None``, auto-computed.
+    constraints : list of callables, optional
+        Outcome constraint callables.
+    eta : float, default=1e-3
+        Temperature for the sigmoid approximation of the constraint indicator.
+    **kwargs
+        Forwarded to :class:`QBatchBoTorchAcquisition`.
+    """
+
+    def __init__(
+        self,
+        *,
+        best_f: Optional[float] = None,
+        constraints: Optional[list[Callable[[torch.Tensor], torch.Tensor]]] = None,
+        eta: float = 1e-3,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+        self._best_f_override = best_f
+        self._constraints = constraints
+        self._eta = eta
+
+    def _build_botorch_acquisition(self) -> Any:
+        """Construct the BoTorch qExpectedImprovement object."""
+        assert self._botorch_surrogate is not None
+        best_f = self._resolve_best_f(self._best_f_override)
+        build_kwargs: dict[str, Any] = {
+            "model": self._botorch_surrogate.get_model(),
+            "best_f": -best_f if not self.maximize else best_f,
+            "constraints": self._constraints,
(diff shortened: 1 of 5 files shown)


src/activelearning/acquisition/botorch/botorch_qbatch.py
```diff
@@ -0,0 +1,479 @@
...
+        }
+        if not self.maximize:
+            build_kwargs["posterior_transform"] = ScalarizedPosteriorTransform(
+                weights=torch.tensor([-1.0], dtype=train_X.dtype, device=train_X.device)
+            )
+        return _qKG(**build_kwargs)
+
+
+class QMaxValueEntropy(QBatchBoTorchAcquisition):
+    """Monte Carlo q-Max-Value Entropy Search (qMES).
```
carriepl, line 395: General comment for this file; the class docstrings would benefit from a bit more info regarding the acquisition strategy that they implement. Maybe at least a small description of the strategy, how it works, and a link to additional info (for instance, to the wrapped BoTorch object, if that's where additional info can be found).

tests/acquisition/test_botorch_acquisition.py
```diff
@@ -1092,3 +1195,336 @@ def test_qmfkg_maximize_false(
...
+        )
+
+        acq = QMaxValueEntropy(
+            candidate_set_spec=TrainDataCandidateSetSpec(),
+            num_fantasies=2,
+            num_mv_samples=5,
+            num_y_samples=16,
+        )
+        acq.update(sf_surrogate, sf_obs)
+        self._scores_valid(acq.score(cands))
```
carriepl, line 1276: For this test and the following ones, ideally, the tests should go in a bit more depth to validate that the results are correct. Only checking that the Acquisition produces the right number of non-infinite floats is a very easy test to pass.

If there was an implementation mistake and, under the hood, all the different batch Acquisition classes ignored all arguments provided during their `__init__()` and instead returned random floats in the range [1, 10], then it seems like all the tests from here to line L1508 would still pass.

PR #11, opened 2026-04-06
Features/pending aware selector

olutayo-mila:
# PR: Pending-aware greedy selector and `score_with_pending`

**Base branch:** `features/qbatch-acquisitions`  
**Head branch:** `features/pending-aware-selector`

---

## Summary

Adds pending-aware greedy batch construction: a selector that conditions
each candidate's score on the candidates already selected in the current
round. This avoids redundant selections by accounting for inter-candidate
dependencies via BoTorch's `set_X_pending` mechanism.

Also refactors the `Selector` ABC from `__call__` to an explicit
`select()` method for clarity and consistency.

---

### `score_with_pending()` (`botorch_acquisition.py`)

New method on `BoTorchAcquisitionBase` enabling pending-aware singleton
scoring:

```python
scores = acquisition.score_with_pending(candidates, pending_candidates)
```

- Sets `X_pending` on the live BoTorch acquisition object via
  `set_X_pending()`, scores, then restores `X_pending = None` via
  `try/finally` (safe even if scoring raises).
- Falls back to `score()` when `pending_candidates` is empty.
- `supports_pending_scoring` property is always `True` for BoTorch
  acquisitions since all BoTorch acquisition functions expose
  `set_X_pending()`.
- Supports `cost_weighting` post-processing, matching the `score()` API.

### New file: `pending_aware_greedy_selector.py`

`PendingAwareGreedyFantasySelector` implements greedy batch construction:

1. At each step, score all remaining candidates conditioned on the
   already-selected set (via `score_with_pending`).
2. Select the highest-scoring feasible candidate.
3. Repeat up to `k` times or until the pool / budget is exhausted.

Supports optional budget-constrained selection when both `cost_fn` and
`round_budget` are provided.

### Selector ABC refactor (`selector.py`)

- Renamed `__call__` → `select()` on the abstract base class.
- Updated `TopKAcquisitionSelector`, `CostAwareSelector`, and all call
  sites (`active_learning.py`, tests).

---

## Design decisions

**Why `score_with_pending` on the acquisition rather than the selector?**  
The pending mechanism is a property of the acquisition function (it
requires `set_X_pending()` on the BoTorch object). Placing it on the
acquisition keeps the selector simple — it just calls
`score_with_pending` in a loop — and allows other selectors or user code
to reuse pending-aware scoring without duplicating the `set_X_pending` /
`try/finally` logic.

**Why `select()` instead of `__call__`?**  
An explicit method name is clearer for a public API. `__call__` obscures
the selector's role when reading `selector(candidates, ...)` vs
`selector.select(candidates, ...)`.

**Why duck-typed `supports_pending_scoring` rather than an ABC method?**  
Only BoTorch acquisitions support `set_X_pending`. Adding it to the
`Acquisition` ABC would force non-BoTorch implementations to stub out a
method they cannot meaningfully implement. The selector checks via
`getattr(acquisition, "supports_pending_scoring", False)`, which cleanly
rejects unsupported acquisitions with an informative error.

---

## Testing

- **`TestScoreWithPending`** (7 tests): empty input, empty pending
  fallback, pre-update error, one score per candidate, `X_pending`
  cleanup after success, `X_pending` cleanup after error,
  `supports_pending_scoring` flag.
- **`TestPendingAwareGreedyFantasySelector`** (16 tests):
  construction validation, input validation, greedy selection order,
  `k ≥ pool` handling, no duplicates, pending set growth verification,
  budget constraints (respects budget, all unaffordable, heterogeneous
  costs).
- Updated existing selector tests for `__call__` → `select()` rename.

**432 tests passing.**

### DIFF
--- a/src/activelearning/acquisition/botorch/botorch_acquisition.py
+++ b/src/activelearning/acquisition/botorch/botorch_acquisition.py
@@ -358,6 +358,117 @@ class BoTorchAcquisitionBase(Acquisition, ABC):
             return raw_scores
         return cost_weighting(raw_scores, cand_list)
 
+    @property
+    def supports_pending_scoring(self) -> bool:
+        """Whether this acquisition supports pending-aware singleton scoring.
+
+        Always ``True`` for BoTorch acquisitions: all BoTorch acquisition
+        functions expose ``set_X_pending()``, which is used by
+        :meth:`score_with_pending` to condition on already-selected candidates
+        without rebuilding the acquisition object.
+        """
+        return True
+
+    def score_with_pending(
+        self,
+        candidates: Iterable[Candidate],
+        pending_candidates: Iterable[Candidate],
+        cost_weighting: Optional[
+            Callable[[list[float], list[Candidate]], list[float]]
+        ] = None,
+    ) -> list[float]:
+        """Score candidates conditioned on a set of pending (already-selected) candidates.
+
+        Intended for pending-aware greedy batch construction: at each step the
+        already-selected candidates are passed as ``pending_candidates`` so the
+        acquisition accounts for their expected information gain before scoring
+        the remainder.
+
+        Internally this calls ``set_X_pending(X_pending)`` on the live BoTorch
+        acquisition object, where ``X_pending`` has shape ``(P, d)`` — the full
+        model-space dimensionality including the fidelity column in multi-fidelity
+        mode. For each candidate ``X`` with shape ``(N, 1, d)``, BoTorch
+        broadcasts the pending tensor and evaluates a q-batch of size ``1 + P``,
+        so each score reflects the marginal value of the candidate *given* the
+        pending observations.
+
+        The pending state is always restored to its previous value via
+        ``try/finally``, so the stored acquisition is left unchanged after the
+        call, regardless of whether scoring raises.
+
+        Falls back to :meth:`score` when ``pending_candidates`` is empty.
+        Returns a constant score of ``1.0`` for every candidate when the
+        acquisition has not yet been coupled to a fitted surrogate, consistent
+        with :meth:`score` behaviour on the first round.
+
+        .. warning::
+            **Entropy-search acquisitions (qMES, qMFMES)** override
+            ``set_X_pending`` to fit a full fantasy GP on the pending points.
+            Each call to this method therefore triggers a GP fantasization step,
+            which can be expensive when selecting large batches with a greedy
+            loop. For cheaper acquisitions (qEI, qNEI, qKG, qUCB, etc.),
+            ``set_X_pending`` is a simple attribute assignment and the overhead
+            is negligible.
+
+        .. note::
+            **Multi-fidelity mode**: ``encode_candidates`` requires every
+            candidate — including those in ``pending_candidates`` — to have a
+            ``fidelity`` value set. Passing candidates without fidelity in MF
+            mode will raise a ``ValueError``.
+
+        Parameters
+        ----------
+        candidates : Iterable[Candidate]
+            Candidates to score.
+        pending_candidates : Iterable[Candidate]
+            Candidates that have already been selected in the current batch but
+            not yet observed. The acquisition will condition on these.
+            In multi-fidelity mode every pending candidate must carry a
+            ``fidelity`` value.
+        cost_weighting : callable, optional
+            If provided, called as ``cost_weighting(raw_scores, candidates)``
+            after scoring.
+
+        Returns
+        -------
+        result : list[float]
+            Acquisition scores in the same order as ``candidates``.
+            All ``1.0`` when called before ``update()``.
+
+        Raises
+        ------
+        ValueError
+            If the surrogate was fitted in multi-fidelity mode and any candidate
+            in ``pending_candidates`` does not carry a fidelity value.
+        """
+        cand_list = list(candidates)
+        if not cand_list:
+            return []
+
+        pending_list = list(pending_candidates)
+        if not pending_list:
+            return self.score(cand_list, cost_weighting)
+
+        if self._botorch_acqf is None:
+            return [1.0] * len(cand_list)
+
+        assert self._botorch_surrogate is not None  # type: ignore[union-attr]
+
+        X = self._botorch_surrogate.encode_candidates(cand_list).unsqueeze(1)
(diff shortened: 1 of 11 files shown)


src/activelearning/selector/pending_aware_greedy_selector.py
```diff
@@ -0,0 +1,131 @@
...
+        spent = 0.0
+
+        while remaining and len(selected) < self.k:
+            if use_budget:
+                costs = cost_fn(remaining)  # type: ignore[misc]
+                feasible_with_costs = [
+                    (cand, cost)
+                    for cand, cost in zip(remaining, costs)
+                    if spent + cost <= round_budget  # type: ignore[operator]
+                ]
```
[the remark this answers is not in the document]
olutayo-mila: Good catch. Fixed 8ffedba889b4257df7d210f998f6212bca560c05.

src/activelearning/acquisition/botorch/botorch_acquisition.py
```diff
@@ -352,6 +352,111 @@ def score(
...
+        assert self._botorch_surrogate is not None  # type: ignore[union-attr]
+
+        X = self._botorch_surrogate.encode_candidates(cand_list).unsqueeze(1)
+        X_pending = self._botorch_surrogate.encode_candidates(pending_list)
+
+        botorch_acqf.set_X_pending(X_pending)
+        try:
+            raw_scores = self._score_encoded(X)
+        finally:
+            botorch_acqf.set_X_pending(None)
```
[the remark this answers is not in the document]
olutayo-mila: Good catch. Fixed 8ffedba889b4257df7d210f998f6212bca560c05.

src/activelearning/selector/pending_aware_greedy_selector.py
```diff
@@ -0,0 +1,133 @@
...
+                feasible_costs = [cost for _, cost in feasible_with_costs]
+            else:
+                feasible = remaining
+                feasible_costs = None
+
+            scores = acquisition.score_with_pending(feasible, list(selected))
+            best_idx = max(range(len(feasible)), key=lambda i: scores[i])
+            best = feasible[best_idx]
+            selected.append(best)
+
```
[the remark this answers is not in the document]
olutayo-mila: Good catch. Fixed: 6c1b7d0b730876e294fcb5635d0fce7ed4068fa8.

src/activelearning/selector/selector.py
```diff
@@ -13,7 +13,7 @@ class Selector(ABC, ALRuntimeMixin):
     """
 
     @abstractmethod
-    def __call__(
+    def select(
```
carriepl, line 16: This change might require a few corresponding changes in the documentation (which, depending on the order in which PRs are merged, may or may not be handled by the automated documentation processes)

src/activelearning/selector/pending_aware_greedy_selector.py
```diff
@@ -0,0 +1,133 @@
...
+        remaining = list(candidates)
+        if not remaining:
+            return []
+
+        selected: list[Candidate] = []
+        spent = 0.0
+
+        while remaining and len(selected) < self.k:
+            if use_budget:
+                costs = cost_fn(remaining)  # type: ignore[misc]
```
carriepl, line 107: Could there be a scenario in which querying the cost for an example could become costly in itself? For very expensive oracles, for instance, where it might therefore make sense to be willing to more accurately estimate labeling cost (even if this estimation becomes a bit more expensive since it's still much much less expensive then running the oracle).

 In a scenario like that, if would be more efficient to get the list of costs once and prune the list as candidates are selected.

tests/selector/test_pending_aware_greedy_selector.py
```diff
@@ -0,0 +1,209 @@
...
+        assert len(result) == 3
+
+    def test_selects_best_first(self) -> None:
+        """First selected candidate is the globally highest scorer."""
+        candidates = make_candidates(4)
+        scores = [1.0, 4.0, 2.0, 3.0]
+        acq = make_acquisition(scores)
+        sel = PendingAwareGreedyFantasySelector(k=1)
+        result = sel.select(candidates, acquisition=acq)
+        assert result[0] == candidates[1]  # score 4.0
```
carriepl, line 113: At this point, might as well increase k and also make sure that the next selected candidates also follow the expector order.

tests/selector/test_pending_aware_greedy_selector.py
```diff
@@ -0,0 +1,209 @@
...
+        candidates = make_candidates(3)
+        acq = make_acquisition([3.0, 2.0, 1.0])
+        sel = PendingAwareGreedyFantasySelector(k=3)
+        with pytest.raises(ValueError, match="negative cost"):
+            sel.select(
+                candidates,
+                acquisition=acq,
+                cost_fn=lambda _: [1.0, -0.5, 1.0],
+                round_budget=5.0,
+            )
```
carriepl, line 209: How about if some item have cost 0? Are they handled correctly?

How about a test where your evaluate with a real acquisition to make sure that if you have super highly scored identical candidates, you do get the expected behavior that only one of them will be selected but not the other one.

tests/acquisition/test_botorch_acquisition.py
```diff
@@ -631,10 +631,116 @@ def test_batch_cost_weighting_not_applied_before_update(
...
+        acq = QExpectedImprovement()
+        acq.update(fitted_surrogate, single_fidelity_observations)
+
+        def raise_on_call(X: Any) -> list[float]:
+            raise RuntimeError("simulated scoring failure")
+
+        acq._score_encoded = raise_on_call  # type: ignore[method-assign]
+        with pytest.raises(RuntimeError, match="simulated scoring failure"):
+            acq.score_with_pending(candidates[1:], [candidates[0]])
+        assert acq._botorch_acqf.X_pending is None  # type: ignore[union-attr]
```
carriepl, line 741: It would be great to have some additional tests. For instance, do you observe the expected adjustement to a candidate's score when the pending set already includes a duplicate of that candidate?
If you have a pending set containing candidate A, and you want to score candidates A, B, C such that dist(A, A) << dist(A, B) << dist(A, C), do you observe a sharper reduction in scores for A than for B than for C if you compare to the setting with an empxy pending set?

PR #12, opened 2026-04-13
Multi-fidelity environment wrapper

alexhernandezgarcia:
This PR initiates the implementation of a wrapper for GFlowNet environments to incorporate the sampling of a fidelity index in the trajectory of any GFlowNet environment.

The multi-fidelity wrapper (MF wrapper) relies on the Composite environments [implemented in the gflownet library](https://github.com/alexhernandezgarcia/gflownet/tree/main/gflownet/envs/composite). In particular, the MF wrapper is a Composite environment consisting of the base environment and a fidelity environment implemented as a [Choice](https://github.com/alexhernandezgarcia/gflownet/blob/main/gflownet/envs/choice.py) environment. This means that the trajectories of the resulting MF wrapper environment will sample the actions to unfold a trajectory of the base environment, as well as the selection of the fidelity index.

This initial implementation defaults to using a [Set](https://github.com/alexhernandezgarcia/gflownet/blob/main/gflownet/envs/composite/setfix.py) as the wrapper, meaning that the fidelity action is not forced to be selected at either the beginning or the end. Instead, it can be chosen at any point in the trajectory. This default implementation corresponds to the wrapper environment `MultiFidelityGFlowNetEnvWrapper`. However, this PR also implements `MultiFidelityGFlowNetEnvWrapperFidFirst` and `MultiFidelityGFlowNetEnvWrapperFidLast`, which are [Stacks](https://github.com/alexhernandezgarcia/gflownet/blob/main/gflownet/envs/composite/stack.py) and enforce that the fidelity is selected before and after, respectively, the actions of the base environment.

This PR does **not** complete the implementation of the GFlowNet sampler into the framework, which is why the base branch is `gflownet_sampler` and not `main`.

### DIFF
--- a/src/activelearning/sampler/gflownet/gflownet_sampler.py
+++ b/src/activelearning/sampler/gflownet/gflownet_sampler.py
@@ -1,10 +1,14 @@
-import torch
+from functools import partial
 from typing import Any, Iterable, Optional
+
+import torch
 from omegaconf import DictConfig
-from activelearning.sampler.sampler import Sampler
-from activelearning.sampler.gflownet.logger_wrapper import (
-    RuntimeGFlowNetLoggerWrapper,
+
+from activelearning.sampler.gflownet.logger_wrapper import RuntimeGFlowNetLoggerWrapper
+from activelearning.sampler.gflownet.multi_fidelity_env_wrapper import (
+    MultiFidelityGFlowNetEnvWrapper,
 )
+from activelearning.sampler.sampler import Sampler
 from activelearning.utils.types import Candidate, Observation
 
 
@@ -29,6 +33,8 @@ class GFlowNetSampler(Sampler):
         Torch device string (e.g. ``"cpu"`` or ``"cuda"``).
     float_precision : int
         Floating-point precision (32 or 64).
+    n_fidelities : int
+        Number of fidelities. By default, 1 (single fidelity).
     """
 
     def __init__(
@@ -37,6 +43,7 @@ class GFlowNetSampler(Sampler):
         conf: DictConfig,
         device: str,
         float_precision: int,
+        n_fidelities: int = 1,
     ) -> None:
         import hydra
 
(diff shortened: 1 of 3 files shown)


src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -45,11 +52,15 @@ def __init__(
         self._gflownet_device = device
         self._gflownet_float_precision = float_precision
 
-        self.env_maker = hydra.utils.instantiate(
+        env_base = hydra.utils.instantiate(
             self.conf.env,
             device=self._gflownet_device,
             float_precision=self._gflownet_float_precision,
-            _partial_=True,
```
pedromsferraz-mila, line 52: Is there a reason to remove `_partial_=True` here?

I think this introduces a bug as `env_base` is now instantiated once and shared across all `env_maker()` calls (lines 65, 116, 204) via the `partial`. For example, the `env` built on line 65 and the `env` built on line 116 inside `_build_agent()` will wrap the exact same `env_base` object, so sampling a trajectory through one will move `env_base.state` and corrupt the other. I've added a test demonstrating this behaviour (that currently fails): 6af8c78573afe26de880a2b3458881a9edd793a0

alexhernandezgarcia: Good catch! I removed it without thinking of the consequences. Thanks for the test!

To be fixed.

alexhernandezgarcia: Fixed in these commits:
- https://github.com/milaforscience/activelearning/pull/12/commits/182f57cdf4f58f46fc7009eb279cae2438239e5e
- https://github.com/milaforscience/activelearning/pull/12/commits/df12f85e21ee669a37c209af105c585d015d5f8a
- https://github.com/milaforscience/activelearning/pull/12/commits/a86e037e61772db7b6269c2217b53888ca83196d

In short, I have not only added `_partial=True` but updated the wrapper env so that it receives a maker instead of an instantiated object.

pedromsferraz-mila: LGTM, thank you for the fix!

PR #13, opened 2026-04-14
Add support to SELFIES molecules with DKL surrogates, xTB oracle, and GFlowNet configs

pedromsferraz-mila:
## Summary

This PR adds a molecule application built around SELFIES strings, Deep Kernel Learning surrogates, and a multi-fidelity xTB oracle for EA/IP discovery.

Main contributions:

- add SELFIES molecule components:
  - `SelfiesTokenizer`
  - `SelfiesTransformerEncoder`
  - `SelfiesKernel`
  - `ExactSelfiesDKLSurrogate`
  - `VariationalSelfiesDKLSurrogate`
  - `XTBIPEAOracle`
- add molecule experiment configs covering:
  - pool-based exact DKL
  - pool-based multi-fidelity exact DKL
  - single-fidelity SELFIES GFlowNet
  - multi-fidelity SELFIES GFlowNet with exact DKL
  - multi-fidelity SELFIES GFlowNet with variational DKL
- document the molecule workflow, including the SELFIES/xTB tutorial and the cost-weighting behavior for pool vs GFlowNet multi-fidelity setups

## Notes

- Pool-based multi-fidelity molecule configs keep cost handling in the selector, while the GFlowNet multi-fidelity configs keep cost handling in the acquisition so the sampler reward is cost-aware.
- Molecule-specific imports remain optional, so non-molecule workflows do not require the molecule extras.

## Validation

- `uv run pytest`
- `uv run pre-commit run --all-files`
- `make docs-build`


src/activelearning/applications/molecules/xtb_oracle.py
```diff
@@ -0,0 +1,678 @@
...
+        }
+
+        fidelity_configs: dict[int, dict[str, Any]] = {
+            fid: {
+                "cost_per_sample": fidelity_costs[fid],
+                "fidelity_confidence": confidences[fid],
+                "score_fn": lambda mol_str, _fid=fid: self._xtb_score(mol_str, _fid),
+            }
+            for fid in fidelity_costs
+        }
```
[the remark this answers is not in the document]
pedromsferraz-mila: Good point, fixed: [e808450](https://github.com/milaforscience/activelearning/pull/13/commits/e8084509eb1af28d6c04046e3048acf530e4b4ae)

src/activelearning/applications/molecules/config.py
```diff
@@ -0,0 +1,233 @@
...
+    )
+
+
+# ---------------------------------------------------------------------------
+# Encoder configs
+# ---------------------------------------------------------------------------
+
+
+class SelfiesTransformerEncoderConfig(BaseModel):
+    """Configuration for :class:`~activelearning.applications.molecules.encoder.SelfiesTransformerEncoder`.
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [b930473](https://github.com/milaforscience/activelearning/pull/13/commits/b9304738a192ea3b8707398273df0f70acca811c)

src/activelearning/applications/molecules/config.py
```diff
@@ -0,0 +1,233 @@
...
+
+    All fields mirror the encoder constructor; changing them via YAML enables
+    easy hyperparameter search (e.g. latent_dim sweep for benchmarking).
+
+    Parameters
+    ----------
+    vocab : list[str]
+        SELFIES alphabet.  Defaults to :data:`~activelearning.applications.molecules.tokenizer.SELFIES_VOCAB_SMALL`.
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: [b930473](https://github.com/milaforscience/activelearning/pull/13/commits/b9304738a192ea3b8707398273df0f70acca811c)

config/molecules/exact_multi_fidelity.yaml
```diff
@@ -0,0 +1,83 @@
+# Molecule IP/EA optimisation with exact SELFIES DKL + MF-MES over three xTB fidelities.
+#
+# This extends the single-fidelity pool example by:
+# - keeping the exact SELFIES DKL surrogate,
+# - exposing xTB fidelities 1 / 2 / 3 in the pool sampler,
+# - using MF-MES to score candidates,
+# - and using cost-aware selection so cheap fidelities are preferred when informative.
+#   In this pool-based config, cost handling lives in the selector rather than
+#   the acquisition to avoid penalizing cost twice.
```
carriepl: At the moment, I don't understand this comment. Maybe I'll understand it later in the review but there might be value in adding a bit more info here.

docs/extension-guide/acquisition.md
```diff
@@ -103,9 +103,17 @@ state, not scoring. Predict inside `score()` or `score_batches()` instead.
...
-not apply cost weighting twice.
+**Two weighting layers** — `cost_weighting` in `score()` is a caller-side
+post-processing hook. It is distinct from acquisition-level weighting baked
+into the acquisition object during `update()` (for example BoTorch
+multi-fidelity `cost_aware_utility`). If both are used, both penalties apply.
+
+**BoTorch MF cost utilities are baked at `update()` time** — when a BoTorch
+multi-fidelity acquisition is configured with `cost_aware_utility`, that
+utility is wired into the internal BoTorch acquisition object during
+`update()`. Later `score()` calls do not toggle it per consumer: samplers and
```
carriepl: This last sentence is a bit confusing. I had to read it a few times and really dissect it to get what it means. I think that you could expand on this a bit to make it simpler to understand for newcomers.

carriepl: I'd expand my original comment to also include the edits in sampler.md and selector.md.

Having read all of them a couple of times, I think that I get the jist of what is being said : 
- Some acquisition instances might (depending on their implentation and config) consider labeling cost in determining acquisition scores. This is the case for BoTorch Multi-Fidelity acquisition classes.
- The Acquisition class' score() function allows users of the function to provide a callable through the 'cost_weighting' argument. This callable can be used as a post-processing step to adjust candidates' acquisition scores based on their labeling cost. Cost-aware selector will use this mechanism to impose a penalty for acquisition cost.
- These two mechanisms are independant. If both are used (cost-aware acquisition + cost-aware selector), both penalties will be applied during the selection phase.
- Which mechanism to use depends on the intended target. Use cost-aware acquisition to impact both the Sampler (through the surrogate) and the Selector. Or use cost-aware selector to impact only the Selector.

I have trouble saying exactly what it is that I find unclear in the text (sorry about that) but I had to read them multiple times to get what they say. I think it might because the text is really really concise. Expanding a bit on the content, giving an example, etc. might make the whole thing less dense and easier to understand.

src/activelearning/acquisition/cost_utility.py
```diff
@@ -0,0 +1,60 @@
...
+    """Divide scores by cost and add an optional fixed offset.
+
+    Parameters
+    ----------
+    values : torch.Tensor
+        Acquisition scores to rescale.
+    costs : torch.Tensor
+        Positive per-item costs aligned with ``values``.
+    fixed_cost : float or torch.Tensor, default=0.0
+        Optional additive offset applied after inverse-cost scaling.
```
carriepl, line 24: I would suggest to rename this last variable. The fact that the method is called "scale by cost" and that this argument is called "fixed_cost" makes it seem like this is something that will be used for scaling, not for an offset. It suggests that the method implementation will look like this instead : 
`return values / (costs + fixed cost)`

src/activelearning/acquisition/cost_utility.py
```diff
@@ -0,0 +1,60 @@
...
+    if len(scores) != len(costs):
+        raise ValueError(
+            "Scores and costs must have the same length for inverse-cost scaling."
+        )
+    scaled = scale_by_cost(
+        torch.tensor(scores, dtype=torch.float64),
+        torch.tensor(costs, dtype=torch.float64),
+        fixed_cost=fixed_cost,
+    )
+    return scaled.tolist()
```
carriepl: Is the only difference between this function and the previous one the fact that this one operates on sequences and the other one on tensors?

I'm asking because the fact that one method contains the name "scale_" and the other "normalize_" looks like an intentional difference meant to communicate information (notably, that the processing isn't the same because, otherwise, why change this word between the two methods).

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
+# **Molecule Discovery with SELFIES and xTB**
+
+This tutorial assumes you are already comfortable with the previous tutorials: running experiments from YAML, composing logger overlays, interpreting multi-fidelity budgets, and configuring the GFlowNet sampler. We will focus on what changes when the search space is **molecules** rather than real-valued vectors.
+
+The molecule examples follow the same budget-constrained active-learning loop introduced earlier, but the candidate \(x\) is now a molecular string. The objective is a molecular property such as **electron affinity** (EA) or **ionisation potential** (IP) computed by xTB, and the sampler proposes SELFIES strings instead of points in a box.
```
carriepl: At this point, I think it's worth saying very roughly what xTB is. Is it an algorithm? A library? Etc.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+
+| Config | Sampler | Surrogate | Acquisition | Fidelity setting | Purpose |
+|--------|---------|-----------|-------------|------------------|---------|
+| `config/molecules/exact.yaml` | Pool file | Exact SELFIES DKL | UCB | pool fidelity `1` only | Stage 1: smallest pool-based baseline |
+| `config/molecules/exact_multi_fidelity.yaml` | Pool file | Exact SELFIES DKL | MF-MES + `CostAwareSelector` | pool fidelities `1 / 2 / 3` | Stage 2: same pool setup with multi-fidelity scoring |
+| `config/molecules/gflownet_exact.yaml` | SELFIES GFlowNet | Exact SELFIES DKL | UCB | fixed fidelity `1` | Stage 3: swap the pool sampler for a GFlowNet |
+| `config/molecules/gflownet_exact_multi_fidelity.yaml` | SELFIES GFlowNet | Exact SELFIES DKL | MF-MES with cost utility | learned fidelity `1 / 2 / 3` | Stage 4: let the GFlowNet learn molecule-fidelity pairs |
+| `config/molecules/gflownet_variational_multi_fidelity.yaml` | SELFIES GFlowNet | Variational SELFIES DKL | MF-MES with cost utility | learned fidelity `1 / 2 / 3` | Stage 5: keep the MF GFlowNet and swap in the scalable variational surrogate |
```
carriepl, line 38: This table contains a few words that I don't think have been defined before UCB, DKL, ...

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+
+!!! note "Small defaults for fast checks"
+    These examples are tuned to be runnable tutorial setups, not fully optimized molecule-discovery runs. The short command overrides below keep the active-learning budget small enough for a quick functional check, and the checked-in GFlowNet examples also use relatively short training schedules in the exact-surrogate stages so you can verify the full loop quickly. For better learning, increase both the oracle budget so the surrogate sees more observations and the GFlowNet optimization steps so the policy can better approximate reward-proportional sampling.
+
+## **What are SELFIES?**
+
+[SELFIES](https://arxiv.org/abs/1905.13741) (**Self-Referencing Embedded Strings**) are a string representation for molecules. Like SMILES, they encode molecular graphs as text. Unlike SMILES, SELFIES are designed so that generated token sequences decode into chemically valid molecular graphs under the SELFIES grammar.
```
carriepl: "Unlike SMILES, SELFIES are designed so that generated token sequences decode into chemically valid molecular graphs under the SELFIES grammar."

Is this intended to say that **all** sequences that are valid under the SELFIES grammar will decode into chemically valid molecular graphs? If so, maybe make that a bit more explicit.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+
+That property is useful for active learning: a sampler can operate over discrete tokens without constantly producing invalid molecules. In this framework:
+
+1. A candidate molecule is stored as a SELFIES string such as `[C][=C][O]`.
+2. The tokenizer maps each SELFIES token to an integer ID, adds special tokens such as `[CLS]` and `[EOS]`, and pads to a fixed length.
+3. The Transformer encoder embeds the token sequence and pools it into one latent molecule vector.
+4. The DKL surrogate fits a GP on those latent vectors and provides posterior predictions to the acquisition function.
+5. The oracle decodes SELFIES to a molecular graph before constructing a 3-D geometry for xTB.
+
+!!! info "Why not optimise SMILES directly?"
```
carriepl: These is a lot of redundancy between this and the paragraph above where the advantage of selfies is discussed. I think it might flow better if this information is aggregated into a single instance instead of two.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+| `config/molecules/exact_multi_fidelity.yaml` | Pool file | Exact SELFIES DKL | MF-MES + `CostAwareSelector` | pool fidelities `1 / 2 / 3` | Stage 2: same pool setup with multi-fidelity scoring |
+| `config/molecules/gflownet_exact.yaml` | SELFIES GFlowNet | Exact SELFIES DKL | UCB | fixed fidelity `1` | Stage 3: swap the pool sampler for a GFlowNet |
+| `config/molecules/gflownet_exact_multi_fidelity.yaml` | SELFIES GFlowNet | Exact SELFIES DKL | MF-MES with cost utility | learned fidelity `1 / 2 / 3` | Stage 4: let the GFlowNet learn molecule-fidelity pairs |
+| `config/molecules/gflownet_variational_multi_fidelity.yaml` | SELFIES GFlowNet | Variational SELFIES DKL | MF-MES with cost utility | learned fidelity `1 / 2 / 3` | Stage 5: keep the MF GFlowNet and swap in the scalable variational surrogate |
+
+!!! tip
+    The easiest way to understand the molecule examples is to work through them in order: start with `exact.yaml`, then add multi-fidelity structure with `exact_multi_fidelity.yaml`, then switch samplers with `gflownet_exact.yaml`, and only introduce the variational surrogate in the last step with `gflownet_variational_multi_fidelity.yaml`.
```
carriepl: In a sense, there are 3 places where the documentation repeats that these are meant as a progression of increasing complexity : before the table, in the "purpose" column of the table, and this paragraph. This 'tip' might not be needed and/or could be folded into the paragraph before the table.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+The built-in [`XTBIPEAOracle`](../reference/activelearning/applications/molecules/xtb_oracle/#activelearning.applications.molecules.xtb_oracle.XTBIPEAOracle) computes one of two objectives:
+
+| Task | Meaning | Intuition |
+|------|---------|-----------|
+| `ea` | [Electron affinity](https://en.wikipedia.org/wiki/Electron_affinity) | Energy change when the molecule accepts an electron. |
+| `ip` | [Ionisation potential](https://en.wikipedia.org/wiki/Ionization_energy) | Energy required to remove an electron. |
+
+The active-learning loop is formulated as a maximization problem. EA is already
+in that form, while IP runs should negate the physical IP so MES maximizes
+`-IP` rather than raw IP.
```
carriepl: Is it useful to describe here how to do this?

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+
+The three xTB fidelities trade cost for accuracy:
+
+| Fidelity | What happens | Cost in the provided configs |
+|----------|--------------|------------------------------|
+| `1` | RDKit/MMFF geometry, then vertical IP/EA with xTB. | `1.0` |
+| `2` | xTB-optimise the neutral geometry, then compute vertical IP/EA. | `3.5` |
+| `3` | Optimise neutral and ionic geometries, then compute adiabatic IP/EA. | `7.0` |
```
carriepl: Personally, I wouldn't be able to understand the difference between how these fidelity levels work here.  Is this tutorial intended for a chemist who instantly understand all of this, or a CS person who probably wouldn't?

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+
+The oracle also controls how many RDKit conformers are generated before xTB
+starts. `oracle.num_conformers` is the global/default count, and
+`oracle.per_fidelity_num_conformers` can override that default for specific
+fidelities. This only changes the RDKit starting-geometry search; it does not
+change the definition of the xTB fidelity itself.
```
carriepl: Some thing as my comment above

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+uv sync --extra molecules --extra aim
+```
+
+The pool-based configs read one SELFIES string per line from:
+
+```text
+config/data/molecules.txt
+```
+
+You can replace that file with your own pool later. For now, keep the checked-in pool so the config paths work unchanged.
```
carriepl: "provided" might feel more natural than "checked in"

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+
+You should see the same round-level fields as in the synthetic tutorials, plus a figure acknowledgement from the console logger:
+
+```text
+[Figure] 'xtb_ea_query_molecules' (not rendered in console)
+[Step 1] round=1 | num_new_samples=5 | round_cost=5.0000 | total_cost=5.0000 | budget_remaining=0.0000
+Done. Rounds: 1 | Total cost: 5.0000
+```
+
+The exact numbers depend on which candidate was selected and whether xTB succeeds for that molecule. Invalid molecules or failed xTB calculations are recorded as `NaN` and filtered before surrogate fitting.
```
carriepl: For the invalid molecules : How do we get invalid molecules if the SELFIES aims to fix this?

For the failed xTB calculations : What are the reasons for xTB failing? Is it only because there's something wrong with the molecule? Is it issues with the library itself? Etc.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+
+Once the plumbing is working, run the full single-fidelity exact-DKL config:
```
carriepl: This is a pretty small nitpick but I would suggest to replace "plumbing" by something else.

It's not super descriptive and feels a bit out of place compared to the tone of the rest of the documentation.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+uv run activelearning config/molecules/exact.yaml
+```
+
+Then move to the multi-fidelity pool version, which keeps the exact SELFIES DKL surrogate but exposes the full xTB fidelity ladder and uses MF-MES with cost-aware selection:
+
+```sh
+uv run activelearning config/molecules/exact_multi_fidelity.yaml
+```
+
+This second stage is the first place where the active-learning loop has to decide how much accuracy is worth paying for. The acquisition scores candidates with respect to the target high-fidelity objective, while the selector divides by query cost so cheap fidelities are favored when they carry similar information. In this pool-based setup, that cost penalty is applied only in the selector; the GFlowNet multi-fidelity configs instead keep the cost model inside the acquisition because the reward itself must become cost-aware.
```
carriepl: I struggled a bit with the last sentence until I read it a few times. By this point in the tutorial, I had forgotten the exact specifications of each experiment setup. I thought that this one had a GFN sampler so the sentence made no sense to me (it simulatenously said that the cost penalty was selector-only and in the acquisition).

I think it might be a bit less confusing to talk less about GFN here since there is none involved at this point. To say how the desired effect is achieve through the use of a cost-aware selector.

And then in later sections (which I haven't reviewed yet so maybe this is already the case), when you start using the GFN for MF, then you can explain that you don't use the cost-aware selector (as done here) since you want the sampler (and thus the reward for its training) to be cost aware.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+
+    ```sh
+    uv run activelearning config/molecules/exact.yaml \
+      oracle.task=ip \
+      logger.run_name=molecules-dkl-exact-ip
+    ```
+
+## **4. Run the single-fidelity SELFIES GFlowNet sampler**
+
+The single-fidelity GFlowNet molecule configs replace the finite pool sampler with a SELFIES sequence environment and stamp every sampled molecule with `fixed_fidelity: 1`:
```
carriepl: I would rephrase slightly since this seems to be implying that it is setting some sample/candidate/molecule/etc's "fixed_fidelity" attribute to have the value 1.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+For the full small tutorial run, drop the overrides:
+
+```sh
+uv run activelearning config/molecules/gflownet_exact.yaml
+```
+
+!!! note "Why keep the single-fidelity GFlowNet config?"
+    It is the cheapest molecule GFlowNet example in the repository and is convenient for checking that the SELFIES environment, xTB oracle, and GFlowNet training loop all run end-to-end before spending budget on multi-fidelity experiments.
```
carriepl: I would maybe leave out the xTB oracle since it has been used in the previous examples and so it seems like this SF-GFN experiment might not have the purpose of testing that oracle.

docs/tutorials/molecule_discovery.md
```diff
@@ -0,0 +1,311 @@
...
+        env:
+            _target_: gflownet.envs.sequences.selfies.Selfies
+            env_id: selfies
+            id: selfies
+            max_length: 8
+```
+
+In this setup, the GFlowNet is not limited to a fixed oracle level. It learns a policy over molecule-fidelity pairs, and the MF-MES acquisition rewards candidates that are informative about the target high-fidelity objective while the cost-aware utility discounts unnecessarily expensive queries. That is one of the main reasons GFlowNets are attractive in the multi-fidelity setting: the same policy can discover promising molecule structures and learn when a cheap xTB evaluation is enough versus when it is worth paying for a higher-fidelity query.
+
+The smaller exact-GP multi-fidelity example is:
```
carriepl: here smaller means "no GFN"? or there are other things that change between the two?

src/activelearning/dataset/list_dataset.py
```diff
@@ -14,15 +36,23 @@ def __init__(self) -> None:
...
+            obs for obs in observations if _has_finite_scalar_target(obs)
+        ]
+        dropped = len(observations) - len(valid_observations)
+        if dropped > 0:
+            logger.warning(
+                "Dropping %d observations with non-finite scalar targets.", dropped
+            )
         self._latest_start_idx = len(self._records)
-        self._records.extend(observations)
+        self._records.extend(valid_observations)
```
carriepl: I'm wondering if this filtering should be responsability of the dataset or of some other part of the code.

This code being here says that there are reasonable circomstances when an oracle might return an infinite value but no reasonable circomstances where such label should make it into the dataset. Do you have an example of a scenario like that? And is there a reason why individual infinite scalars would be prohibited but not lists of infinite scalars?

I'm just wondering if this is potentially an oracle problem or a sampler problem, rather than a dataset problem.

src/activelearning/sampler/gflownet/utils.py
```diff
@@ -10,62 +11,83 @@
...
+        try:
+            value = value.tolist()
+        except TypeError:
+            pass
+    if _is_non_string_sequence(value):
+        return tuple(value)
+    return value
+
+
+def _single_fidelity_proxy_values(proxy_coords: Any) -> list[Any]:
```
carriepl: In this function and the next, why call them "coords"? Especially since the funcction name and the docstring call them values. Does "coords" mean something specific here or is it just a synonym for "values"?

carriepl: Addendum : the method names are also inconsistent, sometimes calling them values and sometimes calling them states.

src/activelearning/sampler/hypercube_sampler.py
```diff
@@ -9,12 +9,15 @@
 
 
 class HypercubeSampler(Sampler):
-    """Generates candidates by sampling from a bounded hypercube.
+    """Generate fresh candidates from a bounded hypercube.
```
carriepl: Why add the word "fresh"? Does this mean that this sample ensures that all samples are novel?

src/activelearning/sampler/hypercube_uniform_sampler.py
```diff
@@ -78,16 +79,18 @@ def sample(
             Unused. Present for interface compatibility.
         observations : Optional[Iterable[Observation]]
             Unused. Present for interface compatibility.
+        cost_fn : Optional[Callable[[Sequence[Candidate]], list[float]]]
+            Unused. Present for interface compatibility.
```
carriepl: If unused, would it be a good idea to raise a warning or error if the user tries to provide a value for this?

src/activelearning/applications/molecules/plotting.py
```diff
@@ -0,0 +1,248 @@
...
+        return str(observation.metadata["raw"])
+    return str(candidate.x)
+
+
+def _decode_molecule(molecule: str, mol_repr: str) -> tuple[str | None, str | None]:
+    """Decode a molecule to SMILES, returning an invalid reason on failure."""
+    if mol_repr == "smiles":
+        return molecule, None
+    if mol_repr != "selfies":
+        return None, f"unsupported representation {mol_repr!r}"
```
carriepl: Is the method implemented this way because there is a plan to include selfies in the future? It's just odd to have a method whose doctring says "I support X" but that also accepts arguments so that the caller can specify that they want it to do Y instead of X (which will return a message).

I guess I'm asking : if an option is not supported, why expose argguments that will give the impression that it is supported? Why not have a single input argument `molecule`?

src/activelearning/applications/molecules/plotting.py
```diff
@@ -0,0 +1,248 @@
...
+        return str(candidate.metadata["raw"])
+    if isinstance(observation.x, str):
+        return observation.x
+    if observation.metadata is not None and "raw" in observation.metadata:
+        return str(observation.metadata["raw"])
+    return str(candidate.x)
+
+
+def _decode_molecule(molecule: str, mol_repr: str) -> tuple[str | None, str | None]:
+    """Decode a molecule to SMILES, returning an invalid reason on failure."""
```
carriepl: Migght be worth documenting the known reasons that can lead to a failure here if those are to be expected.

src/activelearning/applications/molecules/plotting.py
```diff
@@ -0,0 +1,248 @@
...
+        smiles = sf.decoder(molecule)
+    except sf.DecoderError as exc:
+        return None, str(exc)
+    if smiles == "":
+        return None, "empty molecule"
+    return smiles, None
+
+
+def _mol_from_smiles(smiles: str | None) -> Chem.Mol | None:
+    """Build an RDKit molecule from SMILES without emitting RDKit diagnostics."""
```
carriepl, line 199: Random place to put this comment but a lot of docstrings here don't follow the format outlined in the documentation. Was this format only for some specific files?

src/activelearning/applications/molecules/dkl_surrogate.py
```diff
@@ -0,0 +1,639 @@
...
+    def _apply_runtime_context(self) -> None:
+        """Apply the currently bound runtime device/dtype to all learnable modules."""
+        self._encoder = self._encoder.to(device=self.device, dtype=self.dtype)
+        if self.model is not None:
+            self.model = self.model.to(device=self.device, dtype=self.dtype)
+        if getattr(self, "_gp_model", None) is not None:
+            self._gp_model = self._gp_model.to(device=self.device, dtype=self.dtype)
+        if getattr(self, "_likelihood", None) is not None:
+            self._likelihood = self._likelihood.to(device=self.device, dtype=self.dtype)
+        if getattr(self, "_botorch_adapter", None) is not None:
```
carriepl: This is a nitpick and doesn't need to be changed but there is a funcion hasattr (https://www.w3schools.com/python/ref_func_hasattr.asp) that can be used here to simplify the code.

tests/applications/molecules/test_dkl_surrogate.py
```diff
@@ -0,0 +1,608 @@
...
+        training_params=TRAINING,
+        num_inducing=8,
+        multi_fidelity=True,
+        target_fidelity=3,
+    )
+    surrogate.set_fidelity_confidences(FIDELITY_CONFIDENCES)
+    return surrogate
+
+
+class TestVariationalSelfiesDKLSurrogate:
```
carriepl, line 247: General comment about this class and the previous one; there are many test to validate the results but it's mostly ensuring that the right keys are present and that the right dtype is used. Though it would require a bit more effort, it might be good, if possible, to have a few tests to validate the correctness of the outputs (beyond validating that the STD is above 0.0)

tests/applications/molecules/test_selfies_tokenizer.py
```diff
@@ -0,0 +1,102 @@
...
+    def test_encode_selfies_shape(self, tokenizer: SelfiesTokenizer):
+        ids = tokenizer.encode_selfies(BENZENE, max_length=32)
+        assert ids.shape == (32,)
+        assert ids.dtype == torch.long
+
+    def test_encode_selfies_padding(self, tokenizer: SelfiesTokenizer):
+        ids = tokenizer.encode_selfies(BENZENE, max_length=64)
+        # Benzene is short; remaining positions should be padding
+        pad_count = (ids == tokenizer.padding_idx).sum().item()
+        assert pad_count > 0
```
carriepl: This is not a super discriminative test. There could be many possible errors in padding and encoding and you would still get way more than 0 padding positions. Much like the next test which only ensures that the length is higher than 16 for a super long input chain; it will only ever catch very few potential issues.

Since the encoding (from my understanding) is about replacing tokens by their IDs (and adding a few other tokens for special characters and such), you should be able to know the exact length of the proper encoding for all three test strings. Which means that you could ensure, in these tests, that there is exactly the right number of padding and non-padding tokens. Also, even if you don't know the specific ID that each token will be mapped to, you could have tests that ensure that the number of distinct ID values is as expected given the input selfie.

tests/applications/molecules/test_selfies_tokenizer.py
```diff
@@ -0,0 +1,102 @@
...
+        raw = torch.stack([tokenizer.encode_selfies(BENZENE, 32)])
+        out = tokenizer.transform_batch(raw)
+        assert int(out[0, 0]) == tokenizer.cls_idx
+
+    def test_transform_batch_eos_inserted(self, tokenizer: SelfiesTokenizer):
+        raw = torch.stack([tokenizer.encode_selfies(BENZENE, 32)])
+        out = tokenizer.transform_batch(raw)
+        # [EOS] should appear exactly once per sequence
+        eos_count = (out[0] == tokenizer.eos_idx).sum().item()
+        assert eos_count == 1
```
carriepl: Maybe worth ensuring that it is also the case when you provide multiple smiles at once.

tests/selector/test_score_selector.py
```diff
@@ -92,3 +92,23 @@ def test_selection_with_varied_scores():
...
+    selector = TopKAcquisitionSelector(num_samples=2)
+    candidates = [Candidate(x=1), Candidate(x=2), Candidate(x=3)]
+
+    selected = selector(
+        candidates,
+        acquisition=acquisition,
+        cost_fn=lambda cands: [1.0 for _ in cands],
+    )
+
+    assert len(selected) == 2
```
carriepl: Can the test be improved by ensuring that, not only can it request these scores, but also that the right candidates will be selected when doing the normalization?

src/activelearning/applications/molecules/selfies_transformer_encoder.py
```diff
@@ -0,0 +1,319 @@
...
+            Same shape as ``x`` with positional information injected.
+        """
+        return self.dropout(x + self.pe[:, : x.size(1)])
+
+
+class MaskedMeanPool(nn.Module):
+    """Pool token features into one vector per molecules via masked mean + projection.
+
+    The mask excludes padding and ``[EOS]`` so the pooled vector represents the
+    full encoded sequence content, including the leading ``[CLS]`` token.
```
carriepl: Is there a reason for excluding EOS but include CLS if (to to my understanding) both of them will always be present?

carriepl: Additionally, this class only implements a masked mean. There is nothing in the insides of the class that is specific to EOS or CLS token. This means that this paragraph isn't really a comment about what the class is or can do, but rather about how other components will use it. 
For that reason, I think that the content of this comment would be more useful if it were found in the places where this class is being used.

src/activelearning/applications/molecules/selfies_transformer_encoder.py
```diff
@@ -0,0 +1,319 @@
...
+        embed_dim: int = 64,
+        ff_dim: int = 256,
+        num_heads: int = 8,
+        num_layers: int = 8,
+        latent_dim: int = 64,
+        dropout: float = 0.0,
+    ) -> None:
+        super().__init__()
+        self.tokenizer = tokenizer
+        self.max_length = max_length + 2  # +2 for [CLS] and [EOS]
```
carriepl: Sometimes variables named "max_length" includes the two special tokens, sometimes they don't. This gets confusing and will create opportunities for mixups (and, therefore, bugs) in the code. Not just here but in other places in the code that have to handle these sequences (with or without the special tokens).

I would suggest adopting changing the variable naming convention to try and avoid these situations (great blog post that I really like on the subject : https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/). Maybe a variable prefix to identify whether we're talking about the "basic" vs the "augmented" sequence (or some other better suited terms).

src/activelearning/applications/molecules/selfies_transformer_encoder.py
```diff
@@ -0,0 +1,319 @@
...
+        self.embed_dim = embed_dim
+        self.latent_dim = latent_dim
+
+        self.embedding = nn.Embedding(
+            tokenizer.vocab_size,
+            embed_dim,
+            padding_idx=tokenizer.padding_idx,
+        )
+        # +1 extra position buffer for safety (long molecules close to max_length)
```
carriepl: This is not a comment that instills confidence. Is there an issue with the code below where if fails to correctly compute the required length for the positional embedding?

carriepl: This comment gives the impression that there was an original version of the code without the +1 which sometimes gave index errors and the +1 was added as a desperate attempt to fix the error (and it worked, but the person implementing the code doesn't know why)

src/activelearning/applications/molecules/selfies_transformer_encoder.py
```diff
@@ -0,0 +1,319 @@
...
+        valid = (
+            token_batch.ne(tok.padding_idx)
+            & token_batch.ne(tok.cls_idx)
+            & token_batch.ne(tok.eos_idx)
+        )
+        mask = torch.zeros_like(token_batch, dtype=torch.bool)
+        for row in range(token_batch.size(0)):
+            valid_positions = torch.where(valid[row])[0]
+            n_mask = max(1, int(len(valid_positions) * mask_ratio))
+            if len(valid_positions) == 0:
```
carriepl, line 284: Because of the `max(1, ...)`, the actual mask ratio applied to short molecules might be higher than the value in `mask_ratio`. Is this what is desired?

tests/applications/molecules/test_selfies_transformer_encoder.py
```diff
@@ -0,0 +1,166 @@
...
+        pe = PositionalEncoding(embed_dim=16, max_len=64)
+        x = torch.zeros(1, 10, 16)
+        out = pe(x)
+        assert not torch.allclose(out, x)
+
+    def test_odd_embed_dim_supported(self):
+        pe = PositionalEncoding(embed_dim=7, max_len=64)
+        x = torch.randn(2, 20, 7)
+        out = pe(x)
+        assert out.shape == x.shape
```
carriepl, line 72: I would suggest an additional test to ensure that the values are not only not zero, but also correct.

tests/applications/molecules/test_selfies_transformer_encoder.py
```diff
@@ -0,0 +1,166 @@
...
+        out = pool(features, mask)
+        assert torch.isfinite(out).all()
+
+    def test_preserves_double_precision(self):
+        pool = MaskedMeanPool(input_dim=8, output_dim=4).double()
+        features = torch.randn(2, 10, 8, dtype=torch.float64)
+        mask = torch.ones(2, 10, dtype=torch.bool)
+        out = pool(features, mask)
+        assert out.dtype == torch.float64
+
```
carriepl, line 162: Here too, it would be nice to have a test that actually makes sure that the output values are the right ones (does it give the right average for 1 item? multiple items? what if the mask is only 0s?)

tests/applications/molecules/test_selfies_transformer_encoder.py
```diff
@@ -0,0 +1,166 @@
...
+    ):
+        out = encoder(token_batch)
+        assert out.dtype in {torch.float32, torch.float64}
+
+    def test_double_precision(
+        self, encoder: SelfiesTransformerEncoder, token_batch: torch.Tensor
+    ):
+        encoder_d = encoder.double()
+        out = encoder_d(token_batch)
+        assert out.dtype == torch.float64
```
carriepl: If the input dtype is float64, the output dtype should be float64. Cool.

If the input dtype is float32, what is the expected output dtype? The previous test just checks if it is floating point value.

tests/applications/molecules/test_selfies_transformer_encoder.py
```diff
@@ -0,0 +1,166 @@
...
+        loss = encoder.mlm_loss(token_batch, mask_ratio=0.15)
+        assert loss.ndim == 0
+        assert float(loss) >= 0.0
+
+    def test_mlm_loss_zero_mask_ratio_handled(
+        self, encoder: SelfiesTransformerEncoder, token_batch: torch.Tensor
+    ):
+        # mask_ratio=0 → n_mask = max(1, 0) = 1; should still succeed
+        loss = encoder.mlm_loss(token_batch, mask_ratio=0.0)
+        assert torch.isfinite(loss)
```
carriepl: The previous test checks if the loss is higher than 0.0. Why does this test apply a looser criteria? Can the loss be negative here? Even if the provided mask_ratio is 0.0, the implementation still makes sure that at least 1 position should be masked so the loss should follow similar criteria as in the previous test.

tests/applications/molecules/test_selfies_transformer_encoder.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def test_sample_mask_positions_ratio(
+        self, encoder: SelfiesTransformerEncoder, token_batch: torch.Tensor
+    ):
+        mask = encoder.sample_mask_positions(token_batch, mask_ratio=0.5)
+        assert mask.dtype == torch.bool
+        assert mask.shape == token_batch.shape
+
```
carriepl, line 213: The test name implies that this will test the ratio of masked position but it doesn't. It just validates the dtype and shape. You should add some checks to ensure that the samples mask contains the expected number of non-0 values and that the associated positions are fine (since you should, for instance, not sample the EOS tokens).

tests/applications/molecules/test_selfies_transformer_encoder.py
```diff
@@ -0,0 +1,166 @@
...
+    PositionalEncoding,
+    SelfiesTransformerEncoder,
+)
+from activelearning.applications.molecules.selfies_tokenizer import (
+    SELFIES_VOCAB_SMALL,
+    SelfiesTokenizer,
+)
+
+BENZENE = "[C][=C][C][=C][C][=C][Ring1][=Branch1]"
+ALANINE = "[C][C][Branch1][C][N][C][=Branch1][C][=O][O]"
```
carriepl, line 20: Some of the code tested here has sequence-length dependent processing so it would be useful to have additional test cases with very short sequences. Hell, some of the code contains conditions to handle empty sequences so to test that code, there should be an empty sequence in the test cases.

tests/test_optional_molecule_dependencies.py
```diff
@@ -0,0 +1,89 @@
...
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+
+
+REPOSITORY_ROOT = Path(__file__).resolve().parents[1]
+
+
+def _run_with_blocked_molecule_imports(source: str) -> subprocess.CompletedProcess[str]:
```
carriepl: What does the title of the method here refer to? The name seems to suggest but the method runs a comment but does so in a way that the molecule module can't be imported. What part of the code ensures this?

carriepl: Alright so I got later in my review and saw L66 which pretty much confirms that this function is not used to run while blocking molecule imports, it just runs. And it happens to be called and provided with a `source` snippet that blocks molecule imports.
In that case, I think the method should be renamed to reflect what it does, not what other pieces of code call it on.

src/activelearning/applications/molecules/xtb_oracle.py
```diff
@@ -0,0 +1,717 @@
...
+except ImportError as error:  # pragma: no cover - exercised via subprocess test
+    raise missing_molecules_dependency_error("XTBIPEAOracle", error) from error
+
+from activelearning.oracle.multi_fidelity_oracle import MultiFidelityOracle
+from activelearning.utils.types import Candidate, Observation
+from activelearning.applications.molecules.plotting import (
+    build_xtb_query_molecule_figure,
+)
+
+# CODATA 2018 Hartree → eV
```
carriepl: This comment says very little that the line of code doesn't. Maybe expand it a bit.

src/activelearning/applications/molecules/xtb_oracle.py
```diff
@@ -0,0 +1,717 @@
...
+            mol_h,
+            numConfs=conformer_cfg.num_conformers,
+            pruneRmsThresh=conformer_cfg.prune_rms_thresh,
+            maxAttempts=conformer_cfg.max_attempts,
+            useRandomCoords=conformer_cfg.random_coords,
+        )
+
+    num_conformers_generated = mol_h.GetNumConformers()
+    if num_conformers_generated == 0:
+        raise RuntimeError(f"RDKit failed to generate conformers for {smiles!r}")
```
carriepl, line 188: Is this something that is expected for molecules? Also, the docstring doesn't declare the possible exceptions that this method will launch.

src/activelearning/applications/molecules/xtb_oracle.py
```diff
@@ -0,0 +1,717 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _run_xtb(
+    xyz_path: Path,
+    args: Sequence[str],
+    output_path: Path,
+    cwd: Optional[Path] = None,
+) -> subprocess.CompletedProcess:
+    """Run ``xtb <xyz_path> <args...>`` and capture output to a file."""
```
carriepl: Missing sections in docstring

src/activelearning/applications/molecules/xtb_oracle.py
```diff
@@ -0,0 +1,717 @@
...
+    return result
+
+
+def _run_xtb_optimize(
+    xyz_path: Path,
+    gfn_version: int,
+    charge: Optional[int] = None,
+) -> tuple[Path, Path]:
+    """Run ``xtb --opt`` geometry optimisation in ``xyz_path.parent``.
+
```
carriepl, line 304: Missing sections in docstring. I'll stop writing this comment for this file but other functions later in the file might also need some changes to their docstring (arguments, returns, or even the section for exceptions)

src/activelearning/applications/molecules/xtb_oracle.py
```diff
@@ -0,0 +1,717 @@
...
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+
+    def _extract_molecule_string(self, candidate: Candidate) -> str:
+        """Return the molecules string from a candidate."""
+        if isinstance(candidate.x, str):
+            return candidate.x
+        if candidate.metadata is not None and "raw" in candidate.metadata:
+            return str(candidate.metadata["raw"])
```
carriepl: Why is it that the molecule string could be in `.x` or in `.metadata`?

Not necessarily an error, but it sorta tickles me that the code relies on something being available (otherwise a ValueError is raised) but there is no guarantee regarding where it will be available.

src/activelearning/applications/molecules/xtb_oracle.py
```diff
@@ -0,0 +1,717 @@
...
+        self._per_fidelity_num_conformers = self._validate_per_fidelity_num_conformers(
+            per_fidelity_num_conformers, fidelity_costs
+        )
+        self._mol_repr = mol_repr
+        self._negate_score = task == "ip" if negate_score is None else negate_score
+        self.log_molecule_visualizations = log_molecule_visualizations
+        self._molecule_visualization_limit = molecule_visualization_limit
+
+        confidences = self._resolve_fidelity_confidences(
+            fidelity_costs, fidelity_confidences
```
carriepl, line 542: I might be wrong but I don't seen that I have seen code to ensure that only data related to fidelities 1, 2 and/or 3 is received here (which would be a good think to validate since there is a check for this later on at labeling time)

tests/applications/molecules/test_xtb_oracle.py
```diff
@@ -0,0 +1,956 @@
...
+    def test_hartree_to_ev_zero(self):
+        assert hartree_to_ev(0.0) == pytest.approx(0.0)
+
+    def test_hartree_to_ev_negative(self):
+        """Negative energy differences are common in adiabatic IP/EA computations."""
+        assert hartree_to_ev(-1.0) == pytest.approx(-27.2114)
+
+    def test_decode_selfies_to_smiles(self):
+        smiles = _decode_to_smiles(BENZENE_SELFIES, mol_repr="selfies")
+        assert isinstance(smiles, str) and len(smiles) > 0
```
carriepl: You have the expected SMILES string in `BENZENE_SMILES`, can't we compare to the expected string instead of just checking that it is a string with more than 0 characters?

tests/applications/molecules/test_xtb_oracle.py
```diff
@@ -0,0 +1,956 @@
...
+                args=["--gfn", "2", "--vea"],
+                output_path=output_path,
+                cwd=tmp_path,
+            )
+
+
+class TestXTBIPEAOracleConstructionValidation:
+    def test_empty_fidelity_costs_raises(self):
+        with pytest.raises(ValueError, match="fidelity_costs"):
+            XTBIPEAOracle(task="ea", fidelity_costs={})
```
carriepl, line 263: I feel annoying just writing this but... starting here in the file, I started to have a lot of difficulty trying to figure out exactly what each test did. I think that there would be good value in documenting these tests a bit more to make it clearer what exactly they aim to test and how they differ from one another.

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Added a parenthesis explaining broadly what xTB is, let me know if that works for you: f02ae4aab72004504e263105101a1846c6211bba

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, added a "building blocks" paragraph before the table that introduces DKL (with links to ExactSelfiesDKLSurrogate, VariationalSelfiesDKLSurrogate, SelfiesKernel), UCB (linked to UpperConfidenceBound), and MF-MES (linked to QMultiFidelityMaxValueEntropy), so all acronyms in the table are defined and linked to their API docs before the reader hits them: 2c93ed7ffa57d901d7ded36534b0ddd11d989ec3

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Yes, that's indeed what I meant, I rephrased as suggested to make it more clear: e5973916401bc799aa035bb8454913a99966720e

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, rephrased to reduce redundancy: f3197178299c6c42a3647948c444d0dfaa4d2816

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, it's especially redundant after commit 2c93ed7ffa57d901d7ded36534b0ddd11d989ec3. Removed 'tip': ab1de162c52dde43666d0e98a1d84220fa9b3b91

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Now that you mention it, this seems too low-level for this section. Moved to config section: 49b66ddccaff92579b4f729240fd6e28bf783173

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, added a more general description of how these work. My goal is for a CS person to have some idea of what's going on, and for a chemist to be able to quickly understand exactly what's being done. Let me know if the level of detail seems appropriate: 0b13a57e7b07d5357384a80866910e7efdde869a

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, fixed: https://github.com/milaforscience/activelearning/commit/49b66ddccaff92579b4f729240fd6e28bf783173

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point: 0d169920a914aa4b2e7d40b428f0b514aee4e962

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good questions, I've added a note explaining this in more detail, let me know what you think: 96f3feb4c9265e740d26c9ec22a2e8743cb9cdc5

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, fixed: 9833ec94ca9962a4ec1dcfffd96515925442045a

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, I've simplified this part to just mention what's relevant for this example. Let me know what you think: 109115ebd754344d99a82251064e19c197bb73ee

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Note: after the changes from the `gflownet_sampler` branch, we stamp fidelities via `fidelities: [1]` instead of using `fixed_fidelity`.

Rephrased to clarify the mechanism: `fidelities: [1]` restricts the GFlowNet's action space to fidelity-1 only (fidelity selection is removed entirely). Let me know what you think: 195c2498ac16886317ef93cd137921e584993ccd

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, fixed: bc12b40a

docs/tutorials/molecule_discovery.md
pedromsferraz-mila: Good point, there's quite a few changes between the two that weren't mentioned. I've added some explanations: bc12b40a9338056d9e503cbe06a72621f8068e00

src/activelearning/sampler/gflownet/utils.py
pedromsferraz-mila: Good point, fixed on the previous `gflownet_sampler` branch. Here's the [diff](https://github.com/milaforscience/activelearning/pull/13/changes/f4e6af2363680c253e6a16944ba45670c22d5d3f#diff-befb732abf923671a2b216ed1495c2c6d92d425f3c7d95c09aaa6c45ca1a098fL35) for the merge.

src/activelearning/sampler/hypercube_sampler.py
pedromsferraz-mila: I agree this is confusing wording, also fixed on the `gflownet_sampler` branch.

src/activelearning/dataset/list_dataset.py
pedromsferraz-mila: Hm good point, I agree that it's more appropriate to move this closer to the oracle rather than keeping it on the dataset. I've moved this logic to the active learning loop (right after obtaining observations from the oracle, but before adding them to the dataset), and fixed the logic to also include lists/arrays with inf scalars: 1c587d05f53b751f39661a917d4eb410e2449f57

Currently, I made the choice to exclude any inf/nan/None value from the dataset to make it easier for consumers (e.g., otherwise if someone implements a new surrogate model, they'd have to manually call this filter util before fitting the surrogate on the observations). However, in the future it may be worth including observations with failed oracle calls to indicate that we shouldn't sample these points again, or to take into account for the surrogate model, etc. Please let me know your thoughts on this.

config/molecules/exact_multi_fidelity.yaml
pedromsferraz-mila: I modified the descriptions of the configs to make them more self-contained, let me know if this is more clear: 2eaa7c1dcfc5984be91d68edd6d2af366ecb34ef

docs/extension-guide/acquisition.md
pedromsferraz-mila: Thanks for the feedback, I agree that this is a bit confusing... I updated the docs for these pages to explain the cost-weighting mechanisms more explicitly, please let me know if you think it could have more detail / examples, or if this is sufficiently clear: 4d63bfe58f48269424a30343d891024657d2db95

src/activelearning/acquisition/cost_utility.py
pedromsferraz-mila: Good point, fixed: ece42051d2a89f8490c0974bd51dd6814cad8d83

src/activelearning/acquisition/cost_utility.py
pedromsferraz-mila: You're right, this was dead code. I merged the two under the same method: ece42051d2a89f8490c0974bd51dd6814cad8d83

src/activelearning/sampler/hypercube_uniform_sampler.py
pedromsferraz-mila: Good idea, added a util to raise warnings and added calls across methods with unused arguments: be2c672357f9ba3a2165bc2b8dca385b4f7655e0

src/activelearning/applications/molecules/plotting.py
pedromsferraz-mila: You're right, this is indeed confusing. I changed the method to explicitly decode selfies instead: 2e38a05448bdde41f13a1a78cd87cb50a7033685

src/activelearning/applications/molecules/plotting.py
pedromsferraz-mila: Good point, added: 1729491358ec5c410d73f8db91f75fb945aded02

src/activelearning/applications/molecules/plotting.py
pedromsferraz-mila: Good point, in my mind I only required complete docstrings for public methods (i.e., any method that doesn't start with `_`), but now that you mention it I haven't specified this on the documentation. 

What do you think would be best, to adopt this as a standard, or to fully document all methods? I'm open to either option.

src/activelearning/applications/molecules/dkl_surrogate.py
pedromsferraz-mila: Good point, this is cleaner. Fixed: 7f1ee6899a7ef88016618f9e53ecdf93089e2762

tests/applications/molecules/test_dkl_surrogate.py
pedromsferraz-mila: Good point, added some more (simple) correctness tests to the DKL surrogate classes: 2661562b3b7005a8ea67d28cc0e0117a34ce785a

tests/applications/molecules/test_selfies_tokenizer.py
pedromsferraz-mila: Good point, added a parametrized test to do so: 4e181b006669dd1fbb8c008b29052664e73b888b

tests/applications/molecules/test_selfies_tokenizer.py
pedromsferraz-mila: Good point, I added the exact expected distinct count and other tests: 4e181b006669dd1fbb8c008b29052664e73b888b

tests/selector/test_score_selector.py
pedromsferraz-mila: Good point, added: f21ecaebb78f04a230a646b7b78ae81c8acd905d

tests/applications/molecules/test_selfies_transformer_encoder.py
pedromsferraz-mila: Good point, added on `test_sinusoidal_encoding_values:` 4f14f9472c48100b13d7a585ebdeda90c1ca91da

tests/applications/molecules/test_selfies_transformer_encoder.py
pedromsferraz-mila: Good point, added on `test_masked_mean_correct`: https://github.com/milaforscience/activelearning/commit/4f14f9472c48100b13d7a585ebdeda90c1ca91da

tests/applications/molecules/test_selfies_transformer_encoder.py
pedromsferraz-mila: Good point, added a parametrized fixture to test both dtypes: https://github.com/milaforscience/activelearning/commit/4f14f9472c48100b13d7a585ebdeda90c1ca91da

tests/applications/molecules/test_selfies_transformer_encoder.py
pedromsferraz-mila: Good point, added an assertion for loss > 0.0: [4f14f94](https://github.com/milaforscience/activelearning/commit/4f14f9472c48100b13d7a585ebdeda90c1ca91da)

tests/applications/molecules/test_selfies_transformer_encoder.py
pedromsferraz-mila: Added small sequences and an empty string: [4f14f94](https://github.com/milaforscience/activelearning/commit/4f14f9472c48100b13d7a585ebdeda90c1ca91da)

tests/applications/molecules/test_selfies_transformer_encoder.py
pedromsferraz-mila: Good point, added suggested checks on test_sample_mask_positions_ratio: [4f14f94](https://github.com/milaforscience/activelearning/commit/4f14f9472c48100b13d7a585ebdeda90c1ca91da) / [5ef3bae](https://github.com/milaforscience/activelearning/commit/5ef3baecd6965c31222833a428ab62457e84da87)

tests/applications/molecules/test_xtb_oracle.py
pedromsferraz-mila: Good point, fixed: e3f2367f4667be146ffec0ee165e783798b7cfb7

tests/applications/molecules/test_xtb_oracle.py
pedromsferraz-mila: Haha no that's a great point, I've added one-liners to describe each test, let me know if you think they sufficiently clear or could benefit from more descriptive docstrings: 4e33c471eef1fc0910fe7ac0d9a5ec0f051c9399

tests/test_optional_molecule_dependencies.py
pedromsferraz-mila: That's a good point, I've renamed the function to a more descriptive name and added a small docstring: f874915405506daf32963b4b6e8dfbbce55bae8f

src/activelearning/applications/molecules/selfies_transformer_encoder.py
pedromsferraz-mila: Good points. Regarding excluding the EOS token and including the CLS token: since we're using a BERT-style loss and only encoding one sequence at a time (i.e., not multiple molecules), both tokens will always be present. Because of that, I originally thought we only needed one of them. However, I agree there's no harm in keeping both, especially for future-proofing in case those assumptions change. I've updated the code to include both, and I've also moved the comment to the suggested location. 669833407b7a419bcdfb053bb90aa99f4079183a

src/activelearning/applications/molecules/selfies_transformer_encoder.py
pedromsferraz-mila: Good point, that was indeed confusing. I renamed them to use the convention `max_mol_tokens` to refer to the number of molecular (SELFIES) tokens and `max_seq_len` to refer to the maximum sequence length when adding the two special tokens. I've also updated the docstrings, config files and documentation: cd445e658d6834fd63da2aecea3d7cc5e1a5788a

src/activelearning/applications/molecules/selfies_transformer_encoder.py
pedromsferraz-mila: Good catch, that was unnecessary (the `max_seq_len` already accounts for the CLS and EOS tokens). Removed the +1 and comment: 08f067d3e90acb6d63d26667e165c7d24a5d0d3e

src/activelearning/applications/molecules/selfies_transformer_encoder.py
pedromsferraz-mila: Good point. For a 1-token molecule, this could cause all tokens to be masked, leaving the model with `[CLS] [MASK] [EOS]`. This provides no context to predict the `[MASK]`. I've changed the logic to simply skip masking for those sequences instead, which will result in zero loss: 486d77ee71644e3e63191face25c07f254e2fdef

src/activelearning/applications/molecules/xtb_oracle.py
pedromsferraz-mila: Good point, added some more details: 036696e4c524cfd5423f89f21b3f72be58ef1fde

src/activelearning/applications/molecules/xtb_oracle.py
pedromsferraz-mila: Yes, this can happen if RDKit fails to find a feasible solution within `max_attempts` restarts ([RDKit docs](https://www.rdkit.org/docs/source/rdkit.Chem.rdDistGeom.html#rdkit.Chem.rdDistGeom.EmbedMultipleConfs)). Added a comment and docstring to clarify this: de4a2c9a47344ffea0c1e2148a625016b745585d

src/activelearning/applications/molecules/xtb_oracle.py
pedromsferraz-mila: Fixed: 8595c8d10c9981b6367009d9f624f5ae1644a678

src/activelearning/applications/molecules/xtb_oracle.py
pedromsferraz-mila: Good catch, I've added the missing fields throughout the file: 587063b9c884c23f0d7636d00d6a7519685aba58

src/activelearning/applications/molecules/xtb_oracle.py
pedromsferraz-mila: I originally introduced this metadata field with the idea of storing both the molecule embedding and the string on the Candidate object. However, in hindsight, since the molecules are generated by the sampler (which doesn't embed them, that happens on the surrogate), it makes more sense for the Candidate to only hold the molecule string. Leaving the embeddings as part of the surrogate is cleaner and aligns better with the rest of the codebase. Changes: fb69498be4740658d5182d88c34e8792368aa667

src/activelearning/applications/molecules/xtb_oracle.py
pedromsferraz-mila: Good point! Added: d46543114907e3d34ccd605a31035bf25f0d28ff

tests/applications/molecules/test_selfies_transformer_encoder.py
carriepl: Given the change that you made to remove the max(1, ...) in selfies_encoder_transformer, does this mean that this test here should be adjusted? At the very least, there is a comment that refers to the max(1, ...) that should be updated. Also, if no token is masked now, could the loss be 0 (instead of having to be strictly greater than 0)?

pedromsferraz-mila: Good catch. I removed the stale  max(1, ...)  expectations and updated the short-sequence test to assert a zero loss when proportional masking selects no tokens: 14901ab05408a41815a3d9fed9f11f62463b7c1b

config/gflownet/proxy/acquisition.yaml
```diff
@@ -4,3 +4,8 @@ reward_function: power
 reward_min: 1e-8
 reward_function_kwargs:
   beta: 1.0
+
+# MF-GFN paper reward scale: acquisition * reward_rho**round / reward_beta.
+# Defaults preserve raw acquisition rewards.
+reward_beta: 1.0
```
alexhernandezgarcia: I'm afraid this could cause confusion with the `reward_function` and the corresponding parameters that can be passed to the GFlowNet proxy, no?

config/molecules/exact_single_fidelity.yaml
alexhernandezgarcia, line 1: I would call the file `exact_single_fidelity.yaml` to make the single-fidelity aspect explicit, since the library is in general about multi-fidelity active learning.

config/molecules/gflownet_exact_single_fidelity.yaml
alexhernandezgarcia, line 1: I would call the file `gflownet_exact_single_fidelity.yaml` to make the single-fidelity aspect explicit, since the library is in general about multi-fidelity active learning.

src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -95,6 +99,9 @@ def _build_agent(self, acquisition: Acquisition) -> GFlowNetAgent:
         ----------
         acquisition : Acquisition
             Acquisition function used as the GFlowNet reward proxy.
+        cost_fn : callable, optional
```
alexhernandezgarcia, line 103: I find it strange that the cost function needs to be passed to the sampler (GFN) and then set to the proxy.

Shouldn't all this be agnostic to the sampler?

src/activelearning/active_learning.py
```diff
@@ -131,10 +141,22 @@ def active_learning(
...
 
-        # Consume budget and query oracle for new observations, then add to dataset
+        # Consume budget and query oracle for new observations.
+        # Filter out any observations with invalid targets (None, NaN, or infinite)
+        # before adding to the dataset.
+        # Budget is consumed regardless: a failed evaluation still costs compute time.
         budget.consume(total_cost)
         new_observations = oracle.query(selected_samples)
-        dataset.add_observations(new_observations)
+        valid_observations = filter_finite_target_observations(new_observations)
```
alexhernandezgarcia, line 150: Does this call preserve the data in `new_observations`? If so, I'm afraid that this can waste memory, up to duplicating the new observations, right?

alexhernandezgarcia (dismissed):
So much work in here, thanks so much!

I have not reviewed this in any proper depth, but this should be merged.

Please do answer my few questions if you have a moment. 

I have also taken a note to review the documentation, especially the tutorials, soon.

config/gflownet/proxy/acquisition.yaml
pedromsferraz-mila: Good catch, I added these parameters to reproduce the experiments in your paper, but they weren't actually properly implemented. 

I've renamed them to be more clear (and different to the other group of params), and added the intended functionality: 72ffd0b95dbe5a147a279d15217195466e69e1cf

config/molecules/exact_single_fidelity.yaml
pedromsferraz-mila: Good point, fixed: 361c69e11c3c577742e5884999ab0c9349811cda

config/molecules/gflownet_exact_single_fidelity.yaml
pedromsferraz-mila: Good point, fixed: 361c69e11c3c577742e5884999ab0c9349811cda

src/activelearning/sampler/gflownet/gflownet_sampler.py
pedromsferraz-mila: We kept cost weighting outside the acquisition function so samplers can opt into using cost-adjusted rewards during proposal generation, while other components can still consume the raw acquisition scores.

If costs were embedded centrally in the acquisition, every component would be forced to use cost-weighted scores. That may be undesirable for some selectors (especially knapsack-style selectors), which need the raw acquisition and cost separately to apply their own selection logic.

Let me know what you think!

src/activelearning/active_learning.py
pedromsferraz-mila: Yes, it keeps `new_observations` and creates another list, but it does not copy the actual observation data. It only copies references to the valid observations, so the extra memory use is small and temporary.

PR #14, opened 2026-04-24
Add GFlowNet Sampler

pedromsferraz-mila:
# GFlowNet Sampler

This PR integrates [GFlowNet](https://github.com/alexhernandezgarcia/gflownet) as a sampler in the active learning framework.

### GFlowNet Sampler
Adds `GFlowNetSampler`, wrapping the [`alexhernandezgarcia/gflownet`](https://github.com/alexhernandezgarcia/gflownet) library to generate candidates via GFlowNet-guided sampling. The sampler plugs into the existing `Sampler` interface and supports arbitrary GFlowNet environments and acquisition functions.

### Config Parsing (Pydantic + Hydra overloads)
GFlowNet configs are parsed via Pydantic models and support Hydra-style overloads, consistent with the rest of the framework. A set of default configs is provided under `config/sampler/gflownet/`.

### GFlowNetGridSampler
Adds `GFlowNetGridSampler`, a convenience subclass that automatically constructs a discretized grid environment over a bounded continuous domain, making it easy to use GFlowNet sampling in standard bounded optimization tasks without manually configuring the environment.

### Multi-fidelity Wrappers
Builds on the multi-fidelity environment wrappers from [PR #12](https://github.com/milaforscience/activelearning/pull/12) by @alexhernandezgarcia, which introduced `MultiFidelityGFlowNetEnvWrapper`, `MultiFidelityGFlowNetEnvWrapperFidFirst`, and `MultiFidelityGFlowNetEnvWrapperFidLast`. This PR adds support for configuring any of these wrappers through the sampler config.

### Documentation
Adds a GFlowNet sampler tutorial to the docs.


docs/tutorials/gflownet_sampler.md
```diff
@@ -1,6 +1,191 @@
...
-The GFlowNet sampler will train a generative flow network as a proposal mechanism, guided by the current acquisition function as a reward signal. Candidates will be sampled proportional to the acquisition signal, promoting diversity over greedy exploitation.
+Before starting, make sure you have completed [Installation](../getting-started/installation.md).
+
+## **What is a GFlowNet?**
+
+A GFlowNet (Generative Flow Network) is an amortised sampler that learns a stochastic policy $\pi_\theta$ generating objects $x \in \mathcal{X}$ with probability proportional to a non-negative reward $R(x)$:
+
+$$\pi_\theta(x) \propto R(x).$$
+
+Candidates are built step by step along a trajectory $\tau = (s_0 \to s_1 \to \ldots \to x)$, with transitions drawn from a forward policy $P_F(s_{t+1} \mid s_t; \theta)$. Because probability mass is spread over every mode of $R$, the sampler naturally discovers multiple high-reward regions instead of collapsing to one — a property that RL-based alternatives typically lack.
```
carriepl: My reflex would be to nuance this a bit. The GFN is _trained_ to spread probability mass over the different modes, helping the GFN discover multiple high-reward regions.... but there is no guarantee that this will happen.

docs/tutorials/gflownet_sampler.md
```diff
@@ -1,6 +1,191 @@
...
+
+## **GFlowNets in active learning**
+
+In an active learning loop the reward is defined from the **acquisition function**: $R(x, m) = f(\alpha(x, m); \beta)$, where $f$ is a tempered non-negativity transform (exponential or power) controlled by a temperature $\beta$. This step — mapping raw acquisition scores, which may be negative or span many orders of magnitude, to a non-negative reward — is handled by the GFlowNet library's [proxy component](#reward-function) (see [Hernandez-Garcia et al., 2023](https://arxiv.org/abs/2306.11715), Appendix C.3).
+
+Each call to [`sample()`](../reference/activelearning/sampler/sampler/#activelearning.sampler.sampler.Sampler.sample) trains a fresh policy $\pi_\theta$ against the current $R$. As the surrogate sharpens round by round, the GFlowNet adapts automatically: earlier rounds explore broadly, later rounds concentrate near the emerging optima. In multi-fidelity mode the policy jointly samples objects and fidelities $(x, m)$, amortising expensive high-fidelity queries over cheap low-fidelity exploration.
```
carriepl: "As the surrogate sharpens round by round, the GFlowNet adapts automatically: earlier rounds explore broadly, later rounds concentrate near the emerging optima."

I've seen quite a few case where this isn't what happened. Rather, when the surrogate is bad, it sometimes have spurious modes that seem to have really really high objective values so the sampler learns to focus on them way too much. Over the iterations, these spurious modes are explored and removed and the true modes remain.

It depends on the settings of the experiment, although, I would expect this scenario to be much less likely for a GP surrogate than a MLP surrogate.

docs/tutorials/gflownet_sampler.md
```diff
@@ -1,6 +1,191 @@
...
+
+![Branin oracle landscape](../assets/gflownet_oracle.png)
+
+The negated Branin function has three equivalent global minima (red ✕) separated by low-score valleys. This is the landscape the GFlowNet will be trained to sample peaks of — every figure below should be read with reference to these three target regions.
+
+### Sampling the reward landscape
+
+![GFlowNet samples on the target density](../assets/gflownet_samples.png)
+
+The heatmap is the theoretical target $p(x) \propto \exp(\beta \cdot \text{score}(x))$ — the distribution the GFlowNet is trained to match. This example uses a Grid environment, so the GFlowNet builds candidates by stepping along a discrete 2D grid; that is why the white dots appear on a regular lattice rather than at arbitrary positions. Each dot sits on a unique grid cell and its opacity is proportional to how often that cell was visited by the trained policy. A well-trained GFlowNet concentrates bright dots on all three peaks simultaneously; uniform dot brightness across the space indicates under-training or a $\beta$ that is too small; collapse onto a single peak indicates $\beta$ that is too large.
```
carriepl: "indicates $\beta$ that is too large."

->
"indicates *a* $\beta$ that is too large." or "indicates that $\beta$ is too large."

docs/tutorials/gflownet_sampler.md
```diff
@@ -1,6 +1,191 @@
...
+
+### Training convergence ($\log Z$)
+
+![logZ training convergence](../assets/gflownet_logz.png)
+
+The blue curve is the running estimate $\log Z_\theta$; the red dashed line is the exact $\log Z^* = \log \sum_{x, m} R(x, m)$ obtained by enumerating the discrete grid. Convergence of $\log Z_\theta$ to $\log Z^*$ indicates that trajectory balance has reached its fixed point and the policy is sampling proportionally to $R$. A persistent gap usually means insufficient training steps or a poorly chosen `lr_z_mult` — see [Configuration reference](#configuration-reference).
+
+!!! note "Only possible because the grid is finite"
+    Computing $\log Z^*$ exactly requires enumerating every terminal state. For this small Branin grid that is feasible, but for realistic scientific discovery tasks — molecules, proteins, or crystal structures with combinatorially large state spaces — exhaustive enumeration is intractable. In those settings, $\log Z_\theta$ can still be monitored during training as a relative convergence signal, and quality is assessed via the Top-$K$ metrics below.
```
carriepl: I think it would be worth expanding on how someone is supposed to interpret the logZ curves in such a setting. How do you know if it's going OK? What is a signal that something is wrong? Is the convergence of logZ a necessary and/or sufficient condition for anything?

docs/tutorials/gflownet_sampler.md
```diff
@@ -1,6 +1,191 @@
...
+
+### Per-fidelity calibration
+
+![Per-fidelity split-violin calibration](../assets/gflownet_fidelity.png)
+
+One violin per fidelity level. The left half (solid fill) is the $R(x, m)$-weighted theoretical score distribution at that fidelity; the right half (hatched, dashed edge) is the empirical distribution of GFlowNet samples at that fidelity. A calibrated policy produces right halves that mirror the left halves at each fidelity. The target shape itself can differ across fidelities because lower-fidelity oracles are noisier approximations of the high-fidelity one — mismatch at a single fidelity therefore points to that wrapper layer as the bottleneck.
```
carriepl: There is a good explanation of what the graph is but I think there's something missing to explain *why* the graph is here.

Is the goal of this section to advocate for doing this sort of graph as a debugging/monitoring tool? Is it to show the results of a specific experiment? Something else? And depending on the purpose of this section, it will have an effect on how to discuss this graph.

I feel like the section lacks a bit of integration in the greater narrative.

docs/tutorials/gflownet_sampler.md
```diff
@@ -1,6 +1,191 @@
...
+```yaml
+  conf:
+    env:
+      n_dim: 2       # dimensionality of the search space
+      length: 100    # grid cells per dimension — finer = higher resolution
+      max_increment: 1          # max step size per action (default)
+      max_dim_per_action: 1     # dimensions incremented per action (default)
+```
+
+The GFlowNet builds candidates step by step from the source state `[0, 0]`, so trajectory length scales with the grid-index distance between the source and each mode. Setting `max_increment > 1` lets the policy take larger jumps per action — for a 2D Branin grid, `max_increment: 5` cuts the longest possible path from ~200 to ~40 steps and roughly halves training time. Tune `max_increment` together with `length`: a finer grid with larger increments keeps resolution high while keeping trajectories short.
```
carriepl: Since you mention that a user should tune max_increment and length, I think it would be worth explaining a bit more how.

For instance : 
The text says that doing X and Y will make resolution high and trajectories short. Is it desirable for resolution to be high? For trajectories to be short? Always or up to a point? 
Should a user just make `max_increment == length = some huge value` and be done with it (since that will give you super fine resolution and shorter trajectories)?

carriepl: Addendum : AFAIK, saying that "`max_increment: 5` cuts the longest possible path from 200 steps to 40 steps" is wrong. It makes it possible to reach the state furthest from the origin in as little as 40 steps *but* the agent could still pick the action that increments a dimension only by 1 and still take 200 steps to get to the furthest state. 

Unless the environnement somehow makes the actions to increment by 1,2,3 or 4 into terminal actions and only the action to increment by 5 is a non-terminal action but I don't recall that the grid env does this. I seem to recall that the grid env has only one dedicated EOS action.

src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+    def _float_precision(self) -> int:
+        """Return floating-point precision as an integer (32 or 64)."""
+        return 32 if self.dtype == torch.float32 else 64
+
+    def _build_agent(self, acquisition: Any) -> Any:
+        """Build and return a ``GFlowNetAgent`` ready for training.
+
+        Merges runtime device/precision into the config, then calls
+        ``gflownet_from_config``. For multi-fidelity, the env is built as a
+        factory so each copy gets a fresh base env. The acquisition function
```
carriepl: Each copy of what, exactly?

carriepl: Do you mean "through a factory" rather than "as a factory"? I am not sure I understand this part but I'm assuming it refers to the Factory design pattern.

src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+        and runtime logger are injected after construction.
+
+        Parameters
+        ----------
+        acquisition : Any
+            Acquisition function used as the GFlowNet reward proxy.
```
carriepl, line 97: No docstring section for the return value

src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+            env_base_maker = hydra.utils.instantiate(
+                conf.env, device=device, float_precision=fp, _partial_=True
+            )
+            env = build_multi_fidelity_env_wrapper(
+                fidelity_action=self.fidelity_action,
+                env_base_maker=env_base_maker,
+                n_fidelities=self.n_fidelities,
+                float_precision=fp,
+                device=device,
+            )
```
carriepl: This section (L93-L106) is related to the comment that I had trouble understanding above and so I have a bit of trouble understanding it here too. I don't really get why the multi-fidelity setting requires extra work when building the environment. It would be worth adding a comment here and/or clarifying the comment above.

src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+        ``env.states2proxy``; delegates the actual conversion to
+        :func:`~activelearning.sampler.gflownet.utils.proxy_states_to_candidates`.
+
+        Parameters
+        ----------
+        states : tensor or list
+            Terminating states from a trajectory batch.
+        env : GFlowNetEnv
+            The environment used to map states to proxy coordinates.
+        """
```
carriepl, line 177: No docstring section for return value.

src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+        if not isinstance(states, (list, torch.Tensor)) or len(states) == 0:
```
carriepl: I'm wondering about the possibility of `if not isinstance(states, (list, torch.Tensor))` potentially making some errors difficult to investigate.

Are there cases where, if the states is not a list and not a tensor, that you legitimately want to return an empty list? Or would it be more appropriate to raise an error to indicate that there's an implementation issue?

src/activelearning/sampler/gflownet/grid_sampler.py
```diff
@@ -0,0 +1,154 @@
...
+    ----------
+    n_samples : int
+        Number of candidates to generate per :meth:`sample` call.
+    conf : DictConfig
+        Complete GFlowNet configuration tree (env, policy, gflownet, loss,
+        buffer, evaluator, logger, proxy).
+    output_bounds : sequence of (float, float), optional
+        Per-dimension ``(lower, upper)`` bounds to which the grid coordinates
+        are rescaled.  Must have one entry per grid dimension.  If ``None``,
+        the native ``[cell_min, cell_max]`` coordinates are returned as-is.
```
carriepl: Returned where, when? Clearly not here since this is an __init__ but it's not obvious where.

src/activelearning/sampler/gflownet/grid_sampler.py
```diff
@@ -0,0 +1,154 @@
...
+    MultiFidelityGFlowNetEnvWrapperBase,
+)
+from activelearning.utils.types import Candidate
+
+
+class GFlowNetGridSampler(GFlowNetSampler):
+    """GFlowNet sampler that rescales grid coordinates to a target bounded domain.
+
+    The GFlowNet :class:`gflownet.envs.grid.Grid` environment operates in a
+    discrete grid whose cells span ``[cell_min, cell_max]^d``.  This sampler
```
carriepl: This notation is maybe a bit confusing. At this point, cell_min and cell_max have not been defined.

src/activelearning/sampler/gflownet/grid_sampler.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+        grid_min = torch.tensor(float(base_env.cells[0]), dtype=torch.float64)
+        grid_max = torch.tensor(float(base_env.cells[-1]), dtype=torch.float64)
+        return grid_min, grid_max
+
+    def _states_to_candidates(self, states: Any, env: Any) -> list[Candidate]:
+        """Convert GFlowNet states to :class:`~activelearning.utils.types.Candidate` objects.
+
+        If ``output_bounds`` was provided, linearly rescales each coordinate from
+        the grid's native ``[cell_min, cell_max]`` range (read from the built env)
```
carriepl: Is this supposed to be 'cell_min' or 'grid_min'? Because the code below kinda makes it seem like it's rescaling from grid_min and grid_max.

Actually, after going through this class, I am still not sure what cell_min and cell_max are supposed to be.

src/activelearning/sampler/gflownet/config_utils.py
```diff
@@ -0,0 +1,172 @@
...
+    DictConfig
+        Fully composed OmegaConf config ready to be passed to
+        :class:`~activelearning.sampler.gflownet.gflownet_sampler.GFlowNetSampler`.
+    """
+    cfg_dir = _find_config_dir()
+
+    # --- env: base + grid (resolves defaults: [base] in grid.yaml) ---
+    env_cfg = OmegaConf.merge(
+        _load_yaml(cfg_dir / "env" / "base.yaml"),
+        _strip_defaults(_load_yaml(cfg_dir / "env" / "grid.yaml")),
```
carriepl: Does this mean that the grid env is always used?

src/activelearning/sampler/gflownet/logger_wrapper.py
```diff
@@ -0,0 +1,197 @@
...
+        The runtime logger's step is owned by the outer active-learning loop
+        (via ``logger.log_step(num_rounds)``). Calling ``log_step`` here would
+        set the step counter to a GFlowNet training step number (0-N) and corrupt
+        the round numbering. We only track the step locally so ``log_summary``
+        can reference it later.
+        """
+        self._last_step = step
+
+    @staticmethod
+    def _coerce_log_value(value: Any) -> Any:
```
carriepl: coerce? It there a specific reason for this word over the less agressive "convert"?

src/activelearning/sampler/gflownet/multi_fidelity_env_wrapper.py
```diff
@@ -0,0 +1,269 @@
...
+        states_base : List[Any]
+            The states of the base environment in the batch.
+        fidelities : List[int]
+            The fidelity index in each state of the batch.
+        """
+        states_base = []
+        fidelities = []
+        for state in states:
+            states_base.append(self._get_substate(state, self.idx_base_env))
+            fidelities.append(self._get_substate(state, self.idx_fidelity)[0])
```
carriepl: This class requires these attributes to be defined (self.index_base_env, and self.idx_fidelity) but it doesn't ensure that they are defined, meaning that the subclass have to make sure do define them (or get weird cryptic errors if they don't).

I think that the code should be modified to avoid this or, at the very least, this requirement should be explicitely documented in this base class to make sure that contributors don't miss it.

src/activelearning/sampler/gflownet/multi_fidelity_env_wrapper.py
```diff
@@ -0,0 +1,269 @@
...
+        state = self._get_state(state)
+        states_base, fidelities = self.get_states_base_and_fidelities([state])
+        return states_base[0], fidelities[0]
+
+
+class MultiFidelityGFlowNetEnvWrapper(SetFix, MultiFidelityGFlowNetEnvWrapperBase):
+    """Turns any GFlowNet environment into a multi-fidelity environment.
+
+    The wrapper is a SetFix GFlowNet environment whose sub-environments are the
+    base environment and a Choice environment to model the discrete fidelity index.
```
carriepl, line 101: Maybe add a bit of description to explain how this one differs from the other two (related to the timing of the fidelity-choosing actions vs the actions in the other subenv)

src/activelearning/sampler/gflownet/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    tensor and the fidelity scalar respectively.
+
+    For single-fidelity envs, ``states2proxy`` may return:
+    - a 2-D tensor ``[N, D]``
+    - a list of 1-D tensors (stacked internally)
+    - a list of plain sequences
+
+    Parameters
+    ----------
+    proxy_coords : tensor, list, or list of dicts
```
carriepl: Are those always coordinates? I mean, I know that they are in the grid case but in a more general setting (like, say, molecule design), will it always be expressed as coordinates?

src/activelearning/sampler/config.py
```diff
@@ -27,7 +25,77 @@ def build(self) -> Sampler:
...
+    :func:`~activelearning.sampler.gflownet.config_utils.compose_gflownet_conf`.
+    Only experiment-specific overrides need to be provided in ``conf``.
+
+    Parameters
+    ----------
+    type : Literal["GFlowNetSampler"]
+        Discriminator field for the :data:`SamplerConfig` union.
+    n_samples : int
+        Number of candidates to generate per :meth:`~activelearning.sampler.gflownet.gflownet_sampler.GFlowNetSampler.sample` call.
+    n_fidelities : int
```
carriepl: I might be wrong but I think I saw places in the code (can't remember where) where the indices for the different fidelities were somehow arbitrary. If that is the case, wouldn't the GFN need to know the fidelity indices (in case they're not consecutive from 0)? I might be misremembering things here...

src/activelearning/sampler/hypercube_uniform_sampler.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Unlike pool-based samplers, this sampler is generative: it draws fresh
+    random candidates on every ``sample()`` call. This makes it suitable for
+    continuous-domain functions such as ``BraninOracle`` and
+    ``Hartmann6DOracle`` where no finite candidate pool exists.
+
+    Fidelity levels are sampled uniformly at random from ``fidelities`` for
+    each candidate, enabling multi-fidelity exploration without a fixed pool.
+
+    Parameters
```
carriepl: I think that this part of this docstring here should actually go in the docstring of the __init__ method (which is missing)

src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+    Parameters
+    ----------
+    n_samples : int
+        Number of candidates to generate per :meth:`sample` call.
+    conf : DictConfig
+        GFlowNet config (``env``, ``policy``, ``gflownet``, ``loss``,
+        ``buffer``, ``evaluator``, ``logger``, ``proxy``).
+    n_fidelities : int
+        Number of fidelity levels. When > 1 the env is wrapped with a
+        multi-fidelity wrapper chosen by *fidelity_action*.
```
carriepl: I saw other samplers getting the list of fidelities instead of the number of fidelities. If the sample should generate candidates containing the index/id of the fidelity level, should this sampler too not get the list of fidelities?

tests/sampler/gflownet/test_config_utils.py
```diff
@@ -0,0 +1,262 @@
...
+def test_find_config_dir_raises_when_not_found(tmp_path):
+    """_find_config_dir must raise FileNotFoundError when config/gflownet/ cannot be found."""
+    fake_file = tmp_path / "fake_module.py"
+    fake_file.write_text("")
+    # Patch __file__ to point somewhere with no config/gflownet ancestor
+    with patch(
+        "activelearning.sampler.gflownet.config_utils.__file__",
+        str(fake_file),
+    ):
+        # Re-import to pick up the patched __file__ is not needed — call directly
```
carriepl: I'm just a bit confused here by the fact that there is a comment saying that an import is not needed here, right before an import statement. Maybe clarify this part?

tests/sampler/test_hypercube_uniform_sampler.py
```diff
@@ -0,0 +1,151 @@
...
+        HypercubeUniformSampler(bounds=BRANIN_BOUNDS, num_samples=0)
+
+
+def test_empty_bounds_produces_zero_dimensional_candidates():
+    """Empty bounds are allowed and produce 0-dimensional candidate vectors."""
+    sampler = HypercubeUniformSampler(bounds=[], num_samples=5)
+    result = sampler.sample()
+    assert len(result) == 5
+    # Each candidate's x should be 0-dimensional when bounds is empty.
+    assert all(len(candidate.x) == 0 for candidate in result)
```
carriepl: Out of curiosisty, what is the use of supporting this use case?

tests/sampler/gflownet/test_gflownet_sampler.py
```diff
@@ -0,0 +1,328 @@
...
+
+        conf, _ = gflownet_conf_2d
+        with caplog.at_level(
+            logging.WARNING,
+            logger="activelearning.sampler.gflownet.gflownet_sampler",
+        ):
+            GFlowNetSampler(
+                n_samples=3, conf=conf, n_fidelities=1, fidelity_action="any"
+            )
+        assert not caplog.records
```
carriepl, line 123: Why the expected difference in behavior between "any" and ["first", "last"]? Why should there be a warning in one case if there's only 1 fidelity but not in the other?

tests/sampler/gflownet/test_gflownet_sampler.py
```diff
@@ -0,0 +1,328 @@
...
+        candidates = sampler.sample(acquisition=_ConstantAcquisition())
+        assert len(candidates) == 4
+        assert all(isinstance(c, Candidate) for c in candidates)
+
+    def test_sample_returns_2d_coordinates(self, gflownet_conf_2d):
+        """Each candidate has 2 coordinates (matching the 2-D grid env)."""
+        conf, _ = gflownet_conf_2d
+        sampler = GFlowNetSampler(n_samples=3, conf=conf)
+        candidates = sampler.sample(acquisition=_ConstantAcquisition())
+        assert all(len(c.x) == 2 for c in candidates)
```
carriepl, line 277: Not an error per se. But since the comment mentions that these tests require GFN training, they could be merged in the interest of execution speed .

tests/sampler/gflownet/test_gflownet_sampler.py
```diff
@@ -0,0 +1,328 @@
...
+    def test_runtime_logger_end_not_called_by_sampler(self, gflownet_conf_2d):
+        """The sampler must not call logger.end(); that belongs to the AL loop."""
+        conf, _ = gflownet_conf_2d
+        sampler = GFlowNetSampler(n_samples=3, conf=conf)
+        runtime_logger = Mock()
+        sampler.bind_runtime_context(RuntimeContext(logger=runtime_logger))
+
+        sampler.sample(acquisition=_ConstantAcquisition())
+
+        runtime_logger.end.assert_not_called()
```
carriepl, line 299: Same comment about this test and the previous one being so similar that they can be merged.

tests/sampler/gflownet/test_multi_fidelity_env_wrapper.py
```diff
@@ -0,0 +1,295 @@
...
+    env_maker = partial(wrapper_class, env_base_maker=env_base, n_fidelities=2)
+
+    env1 = env_maker()
+    env2 = env_maker()
+
+    assert env1.env_base is not env2.env_base
+
+    initial_state = list(env2.env_base.state)
+    env1.get_random_states(n_states=1)
+    assert env2.env_base.equal(env2.env_base.state, initial_state)
```
carriepl: Maybe I'm missing something (apologies if I am) but isn't this test identical to the previous one except for the parametrize fixture? If so, they could be merged into a single test with 3 possible parametrizations.

tests/sampler/gflownet/test_proxy.py
```diff
@@ -0,0 +1,213 @@
...
+        assert result.shape == (2,)
+
+    def test_acquisition_receives_fidelity(self):
+        acq = _FidelityCapturingAcquisition()
+        self.proxy.set_acquisition(acq)
+        states = _make_mf_states([[0.1, 0.2], [0.3, 0.4]], fidelities=[0, 2])
+        self.proxy(states)
+        assert acq.seen == [0, 2]
+
+    def test_acquisition_receives_base_coords(self):
```
carriepl: Not an error, just an inconsistency but this test evaluates the same thing as "test_acquisition_receives_correct_coords" but the naming scheme is different ("base" vs"correct")

tests/sampler/gflownet/test_multi_fidelity_env_wrapper.py
```diff
@@ -0,0 +1,295 @@
...
+        "env_choice",
+        "env_stack_cube_setgrids",
+        "env_stack_cube_setstacks",
+        "env_set_cubes",
+    ],
+)
+def test__base_envs_initialize_properly(env, request):
+    env = request.getfixturevalue(env)
+    assert True
+
```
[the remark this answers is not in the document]
pedromsferraz-mila: Good point, fixed: b7613259135c537924d613b4cf4ccf3a6e37c3f7

src/activelearning/sampler/gflownet/multi_fidelity_env_wrapper.py
```diff
@@ -0,0 +1,269 @@
...
+
+        This method constructs a dummy batch with the input state and calls
+        ``get_states_base_and_fidelities()``; then returns the first element of the
+        output.
+
+        Parameters
+        ----------
+        states : Any or ``None``
+            A state in environment format. If ``None``, ``self.state`` is used.
+
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: 78452b14b6b2391d2826d2b4e6e22f35745b2ca9

tests/sampler/gflownet/conftest.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def _make_minimal_gflownet_conf(
+    n_train_steps: int = 5,
+    grid_length: int = 5,
+    n_dim: int = 2,
+    cell_min: float = 0.0,
+    cell_max: float = 1.0,
+) -> OmegaConf:
+    """Build a minimal GFlowNet DictConfig suitable for fast unit tests.
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: 653d61746daed70ad272d5a0a72b901dfff9faca

pyproject.toml
```diff
@@ -16,6 +16,8 @@ dependencies = [
     "pydantic>=2.12.5",
     "pydantic-settings>=2.13.1",
     "torch>=2.5.1",
+    "gflownet @ git+https://github.com/alexhernandezgarcia/gflownet.git",
```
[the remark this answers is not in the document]
pedromsferraz-mila: For now, we need to rely on the latest version, but will consider it in the future.

src/activelearning/sampler/gflownet/proxy.py
```diff
@@ -0,0 +1,80 @@
+import torch
+import numpy.typing as npt
+from typing import Any, List, Optional, Union
+from torchtyping import TensorType
+from gflownet.proxy.base import Proxy
+from activelearning.sampler.gflownet.utils import proxy_states_to_candidates
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: cabfc633a7298603e58ca83bed3372a521926fe0

src/activelearning/sampler/gflownet/proxy.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        if self.acquisition is None:
+            raise RuntimeError(
+                "AcquisitionProxy has no acquisition function set. "
+                "Call set_acquisition() before use."
+            )
+
+        candidates = proxy_states_to_candidates(states, self._env)
+        acq_values = self.acquisition.score(candidates)
+        return torch.tensor(acq_values, dtype=self.float, device=self.device)
```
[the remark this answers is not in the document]
pedromsferraz-mila: The implementation is correct as-is, `__call__` returning raw values is exactly what the base class contract requires. Changed the docstring to make clarify: a6cb8a11b1a6d4c3de3cfed8c23c3bd8bfbb6bff

src/activelearning/sampler/hypercube_uniform_sampler.py
```diff
@@ -0,0 +1,103 @@
...
+        self.bounds = bounds
+        self.num_samples = num_samples
+        self.fidelities = list(fidelities) if fidelities is not None else None
+
+        lowers = [b[0] for b in bounds]
+        uppers = [b[1] for b in bounds]
+        self._lower = torch.tensor(lowers, dtype=torch.float64)
+        self._range = torch.tensor(
+            [upper - lower for lower, upper in zip(lowers, uppers)], dtype=torch.float64
+        )
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: c3ae8b3ae82429cfd0a8074f92078bc52412c29a

src/activelearning/sampler/gflownet/multi_fidelity_env_wrapper.py
```diff
@@ -0,0 +1,269 @@
...
+    idx_fidelity : int
+        The sub-environment index corresponding to the fidelity environment. It is set
+        to 1, arbitrarily.
+    """
+
+    def __init__(
+        self,
+        env_base_maker: Type[GFlowNetEnv],
+        n_fidelities: int,
+        **kwargs,
```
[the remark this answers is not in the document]
pedromsferraz-mila: Fixed: dca236fdd82ceb8c3ef1283f197ada425104b901

docs/tutorials/gflownet_sampler.md
pedromsferraz-mila: Good point, I've added some more details to more precisely describe this. Let me know what you think: 3d8c9e26731ad3873d208a0f7b8bf837647fd0fe

docs/tutorials/gflownet_sampler.md
pedromsferraz-mila: Good point, this behaviour doesn't necessarily happen (though with a flat prior the low cost will dominate the cost-weighted acquisition function) and is determined by the acquisition function rather than the GFlowNet. I've rephrased it here to try to make it more clear, let me know what you think: 5c6fed59177ba82a8b126237a5f9583ce448e8d1

docs/tutorials/gflownet_sampler.md
pedromsferraz-mila: Good catch, fixed: fa3eea1bb59bffc6950ed7a7fbd9342a82bff05d

docs/tutorials/gflownet_sampler.md
pedromsferraz-mila: Good point, I've added some discussion on how to interpret logZ, let me know what you think: a38906c27909b1511f63b908ce4fa1a9d41092e7

docs/tutorials/gflownet_sampler.md
pedromsferraz-mila: Good point, the goal here was more to just illustrate what GFlowNet learns, rather than suggesting it as a monitoring tool. I've rephrased it to make it more clear, let me know what you think: 407b27fe3fb3fcd7d6b152c24fb23a1fc0c36092

docs/tutorials/gflownet_sampler.md
pedromsferraz-mila: Good catch, thank you for the comments. I've added more details on the Grid environment and the consequences of setting different values of these parameters, let me know what you think: 0125249a81671019abb4cb494dd9fe40c9362937

pedromsferraz-mila: Addendum: added that we're referring to the 2D grid 8d3cc0368871d15a85b002384bd88531f6e0e9d4

src/activelearning/sampler/gflownet/gflownet_sampler.py
pedromsferraz-mila: That docstring was indeed pretty confusing, I've rephrased it to be more precise about what it's doing, let me know what you think: da1a275ebaeb00f85056301f9e73a201be6d3f2a

src/activelearning/sampler/gflownet/gflownet_sampler.py
pedromsferraz-mila: Good catch, added and updated the type annotations: c8261bd149291ec116e539ffabe98bf9c76864fe

src/activelearning/sampler/gflownet/gflownet_sampler.py
pedromsferraz-mila: I see how this looks confusing, I've moved the logic to build the multi-fidelity env to a separate method with a descriptive docstring (explaining why we need this extra work). Let me know what you think: c2919f66de1fb71baaeefecfc24b9e44a54cdef9 /  a621262d13f16313e9ea45cd87280ae01dd9e481

src/activelearning/sampler/gflownet/utils.py
pedromsferraz-mila: Good point, I've renamed this terminology throughout the files to make it more clear (using the more general term `format` instead of `coordinates`): 2d94de22aaf89f72f3b5e51764a468d1ff0fecae

src/activelearning/sampler/gflownet/gflownet_sampler.py
pedromsferraz-mila: Good catch, fixed: ecdd2f3175fd0c46772e441716b5f9ece1c28e63

src/activelearning/sampler/gflownet/gflownet_sampler.py
pedromsferraz-mila: Good point, now that you mention it I can't think of any case where that would be acceptable. Changed to raise an informative error instead: 3b83cb43f5ad64866677ee6d23ba302343f24a5b

src/activelearning/sampler/gflownet/grid_sampler.py
pedromsferraz-mila: I meant to say they're used as-is (rather than returned, which indeed implies they're returned by some method). Fixed: 659b986793dd59b2d4a45ff18aaf219356f8ecbf

src/activelearning/sampler/gflownet/grid_sampler.py
pedromsferraz-mila: Good point, rephrased to more naturally introduce `cell_min` and `cell_max`: 01f39a533be0ee30b2d4bbce18d42806726eeb25

pedromsferraz-mila: Update: after changing the way I scale the grid env bounds, I've updated the docstring again: 204710f5dc11c4253ef7ffbe25e41f16ba18bbe7

src/activelearning/sampler/gflownet/grid_sampler.py
pedromsferraz-mila: This is indeed very confusing (and while reviewing it myself, I found and fixed a bug...)
The variables `cell_min` and `cell_max` are variables of the `gflownet` library `Grid` env. They determine the min and max coords of a hypercube (with the same min and max values for every dimension). I had implemented `output_bounds` to rescale these to accommodate for different scales for different dims, but the scaling was only applied to the output (which made the acquisition function inputs incorrect for the grid). 

Here's the fixes + updated docstrings and comments explaining the behaviour. Please let me know if this looks more clear: 204710f5dc11c4253ef7ffbe25e41f16ba18bbe7

src/activelearning/sampler/gflownet/config_utils.py
pedromsferraz-mila: That was the default, which users could override via the `conf` field. However, I agree that that's not necessarily a good default on the long run. I've updated it to use the base env as the default and manually specify the `_target_: gflownet.envs.grid.Grid` on the config yaml: ec3dbb484a4827703b3757c0ed5cdef44b640b36

src/activelearning/sampler/gflownet/logger_wrapper.py
pedromsferraz-mila: Not particularly, I agree `convert` makes more sense here. Fixed: ad1f76c31cca1160b395b2119f52f80f044a307d

src/activelearning/sampler/gflownet/multi_fidelity_env_wrapper.py
pedromsferraz-mila: Good point, it'd be good to make this requirement more explicit. I've made these attributes abstract properties in the base class, let me know what you think: 320aa8f000f1e3f4f0e534b9307c849409ee05e2

src/activelearning/sampler/gflownet/multi_fidelity_env_wrapper.py
pedromsferraz-mila: Good point, added: 5ba34c0d6995765273b7692a28adf234c8bf7533

src/activelearning/sampler/config.py
pedromsferraz-mila: Good point, when I implemented this sampler I oversimplified it to just pass `n_fidelities`. I changed the behaviour to follow the same pattern (of receiving a list of fidelities). Internally, the gflownet still samples indices, and those are mapped to the specified fidelities via `proxy_states_to_candidates`. Let me know what you think: 255520f056c83f48cea125eba4e5b90126cd0d3c

src/activelearning/sampler/gflownet/gflownet_sampler.py
pedromsferraz-mila: You're right, that was inconsistent before. Fixed: 255520f056c83f48cea125eba4e5b90126cd0d3c

src/activelearning/sampler/hypercube_uniform_sampler.py
pedromsferraz-mila: Good catch, fixed: 47ff4d026569aa6e71388166ca19aa81d8a8533a

tests/sampler/gflownet/test_config_utils.py
pedromsferraz-mila: Ops, that's outdated or poorly worded, fixed: 22f65388c80ea6ccce4d83fa4e89194cc0286c09

tests/sampler/test_hypercube_uniform_sampler.py
pedromsferraz-mila: Revisiting this now, I agree that there's no use for this. Added a guard instead to ensure bounds is not empty: ac2d32c2af8dcd1edfc1e976743da796a260bf7a

tests/sampler/gflownet/test_gflownet_sampler.py
pedromsferraz-mila: Oh, the only reason we don't raise a warning with "any" is that it's the default. "first"/"last" warn because the user manually specified the flag (which implies multi-fidelity behaviour). If you prefer, I could remove this default and make `fidelity_action` mandatory to specify if `fidelities` is not None. Let me know your thoughts. Meanwhile, I just added a comment to clarify: 256c4af2885b55b44ba71b98c9482dadd9934163

tests/sampler/gflownet/test_gflownet_sampler.py
pedromsferraz-mila: Good point, fixed: 9b8497fe5eb0892e0aa959a1820e9b3ee60a4b92

tests/sampler/gflownet/test_gflownet_sampler.py
pedromsferraz-mila: Good point, fixed: 76afcb99797164960eeddade2a2137c46282156c

tests/sampler/gflownet/test_multi_fidelity_env_wrapper.py
pedromsferraz-mila: You're right, both tests were identical in logic, I merged them into a single test parametrized over `ALL_WRAPPER_CLASSES` (all three variants), removing the redundant `_stack_variants` test: f964fd83f30f478e91c04f5098334802720a3434

tests/sampler/gflownet/test_proxy.py
pedromsferraz-mila: Good catch, fixed: ef0c8a6ef6feefa2ef35e85c49360438cd04a44f

docs/tutorials/gflownet_sampler.md
carriepl: Yup, looks good to me.

docs/tutorials/gflownet_sampler.md
carriepl: I agree with the first part of the changes. However, this part : 
`any shift in sampling behaviour across rounds (e.g., broader early exploration or tighter concentration around optima) is a property of how the acquisition function changes as the surrogate evolves, not of the GFlowNet itself.`
might be a little too strong since the acquisition function does change, but from one round to the next (possibly due to sheer luck), the GFN might end up better trained or worse trained. Therefore, some of the difference in sampling behavior might come from the GFN and not necessarily all of it from the acquisition function.

docs/tutorials/gflownet_sampler.md
carriepl: Looks pretty good to me. I would just change the part that talks about increasing "the action space to 11" to clarify that it's the *size* of the action space that goes up to 11.

docs/tutorials/gflownet_sampler.md
carriepl: Looks pretty good!

src/activelearning/sampler/config.py
carriepl: I think that the change makes a lot of sense.
One suggestion I would make is to rename the variables whose meaning changed with that update. Now, variables called "fidelities" sometime contain fidelity "levels" and sometimes fidelity indices. Enforcing a naming convention would help avoid mixups in the code where a variable containing a fidelity index is mistakenly assumed to contain a fidelity "level" and used as such.

docs/tutorials/gflownet_sampler.md
pedromsferraz-mila: Good point, added this nuance to the docs: 0f1a691c88c399d7ce000b3fd3c31dc256152d29

docs/tutorials/gflownet_sampler.md
pedromsferraz-mila: Good point, added: 6316695b98b4b8aa23fad3849c912df4bb91e38b

src/activelearning/sampler/config.py
pedromsferraz-mila: Good point, added a naming convention for the fidelity indexes in the Choice env: a4c81d98119d1bffea758768ae0d07c7c12bc703

carriepl:
This officially covers all the points that I raised during my review. I don't know if someone else (say, Alex) was scheduled to make a review eventually.

src/activelearning/sampler/gflownet/gflownet_sampler.py
```diff
@@ -0,0 +1,226 @@
...
+        observations: Optional[Iterable[Observation]] = None,
+    ) -> list[Candidate]:
+        """Train a GFlowNet agent and return sampled candidates.
+
+        Parameters
+        ----------
+        acquisition : Optional[Any]
+            Acquisition function used as the GFlowNet reward proxy. Required.
+        observations : Optional[Iterable[Observation]]
+            Unused; reserved for future warm-starting.
```
alexhernandezgarcia, line 205: Just a note: Observations can also be used to improved the training of GFlowNet.

src/activelearning/sampler/gflownet/utils.py
```diff
@@ -0,0 +1,111 @@
...
+            "This likely indicates a bug in the calling code."
+        )
+    if len(states_proxy) == 0:
+        return []
+
+    if isinstance(env, MultiFidelityGFlowNetEnvWrapperBase):
+        idx_base = env.idx_base_env
+        idx_fid = env.idx_fidelity
+        candidates = []
+        for state_proxy in states_proxy:
```
alexhernandezgarcia, line 87: Is there a risk that this for loop that processes states one by one (that is `base.detach().cpu().tolist()`) may be too slow? If this is a bottleneck, we could try to improve it.

alexhernandezgarcia:
About the GFlowNet-related config files (`config/gflownet/*`), is it necessary to have them here? I guess they were copy-pasted from the gflownet repo? I am thinking that this could make it difficult to keep the gflownet and this repo in good sync as we make progress.

alexhernandezgarcia (approved):
Good to merge!

src/activelearning/sampler/gflownet/gflownet_sampler.py
pedromsferraz-mila: For sure, and we do pass them to the GFlowNetSampler.sample method. What I meant is that the current implementation doesn't use the passed Observations (but extensions or subclasses of GFlowNetSampler can easily use them).

src/activelearning/sampler/gflownet/utils.py
pedromsferraz-mila: For CPU-only runs, this is unlikely to be a significant bottleneck, as we're already doing most of the stuff on CPU.

On GPUs, per-state .cpu().tolist() could be very expensive because of repeated synchronization and transfers. However, fixing that properly would require broader changes: states2proxy currently returns Python lists/dicts, and the proxy/acquisition APIs consume Python Candidate objects. 

If we're only using the gflownet library on CPU, I think it's okay to keep it as is (and revisit it only if we want to do an end-to-end GPU batching refactor). Let me know what you think!

